PR #14, opened 2020-01-20
Draft windows logo test 003

yugansan:


lredor:
Hello,
Why do you speak about "Melody Advance" logo? It should be "Capella" logo. No?

PR #26, opened 2020-01-22
[releng] Update Sirius version to 6.3.1-S20200114-10254

lfasani:
Change-Id: I3eb2d4782ce67f9f1ed590040f27d2233d9f98c0

### DIFF
--- a/releng/plugins/org.polarsys.capella.settings/oomph/Capella.setup
+++ b/releng/plugins/org.polarsys.capella.settings/oomph/Capella.setup
@@ -266,7 +266,7 @@
         <repositoryList
             name="Oxygen">
           <repository
-              url="http://download.eclipse.org/sirius/updates/stable/6.1.3-S20190506-061916/oxygen"/>
+              url="http://download.eclipse.org/sirius/updates/stable/6.3.1-S20200114-102549/oxygen"/>
         </repositoryList>
         <repositoryList
             name="2019-06">
--- a/releng/plugins/org.polarsys.capella.targets/full/capella.target-definition.targetplatform
+++ b/releng/plugins/org.polarsys.capella.targets/full/capella.target-definition.targetplatform
@@ -38,7 +38,7 @@ location eclipse "http://download.eclipse.org/releases/2019-06" {
 	org.eclipse.mylyn.help.ui
 }
 
-location sirius "https://download.eclipse.org/sirius/updates/milestones/6.3.0rc1/2019-06/" {
+location sirius "https://download.eclipse.org/sirius/updates/stable/6.3.1-S20200114-102549/2019-06/" {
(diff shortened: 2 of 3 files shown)


sandupostaru:
9 tests were failing: https://ci.eclipse.org/capella/job/capella-product/view/change-requests/job/PR-26/2/

I did a rebase and waiting for the current results: https://ci.eclipse.org/capella/job/capella-product/view/change-requests/job/PR-26/4/

@lfasani Any changes related to drag and drop in the new Sirius version?

lfasani:
I did not see change around drag and drop. Perhaps it could come from the Sirius preference initialization but i don't think so. https://git.eclipse.org/c/sirius/org.eclipse.sirius.git/commit/?id=5fb7368b4bb5ee3386765715f31bb2a8368c80b2
Yet the real change in Capella about preference will be done with https://github.com/eclipse/capella/pull/43

sandupostaru:
I rebased https://github.com/eclipse/capella/pull/43 on this PR, so we have both the Sirius TP updated and the Capella preferences commit. **The same tests are failing** as you can see here https://ci.eclipse.org/capella/job/capella-product/view/change-requests/job/PR-43/4/#showFailuresLink

I looked at some of the failing tests and I found that the reason for the test failure is the `SiriusPreferencesKeys.PREF_AUTO_REFRESH` preference.

Currently in the master branch when we run the tools tests, the  `SiriusPreferencesKeys.PREF_AUTO_REFRESH`  preference **returns false**, in this PR the same preference **returns true**.

**I do not understand the origin of this change, and can only assume it comes from Sirius, since it is the only thing we change in this commit, can you please confirm?**

The implication is that since the preference is true, the 
`org.eclipse.sirius.tools.api.ui.RefreshEditorsPrecommitListener.localChangesAboutToCommit(Collection<Notification>)`
now returns a `RefreshRepresentationsCommand` in 
`org.eclipse.sirius.business.internal.session.SessionEventBrokerImpl.transactionAboutToCommit(ResourceSetChangeEvent)` line 207

Since a refresh is performed now, and not before, the tests expect to find some elements in the drop container which are no longer there because they are reordered in the refresh.

sandupostaru:
I fixed the problem :+1: 

The issue was that the `SiriusPreferencesKeys.PREF_AUTO_REFRESH` is no longer present in the `SiriusEditPlugin.getPlugin().getPreferenceStore()` but in the `SiriusEditPlugin.getPlugin().getCorePreferenceStore()`

All tests are ok now https://ci.eclipse.org/capella/job/capella-product/job/PR-26/6/testReport/

lfasani:
ok good!
However, the tests relying on preference value may be broken for good reasons with the next commit https://github.com/eclipse/capella/pull/43 because there are many fixes in Capella about how to initialize(override) Sirius preferences.

PR #43, opened 2020-01-27
[559169] Fix Sirius preferences initialization

lfasani:
* Initialize preferences in default scope instead of instance scope.
* For MigrationCommandLine force Sirius preferences in the instance
scope

Bug: 559169
Change-Id: I81feb62961593847d6ca9ba2475e5a578b140f4a


common/plugins/org.polarsys.capella.common.platform.sirius.customization/src/org/polarsys/capella/common/platform/sirius/customisation/SiriusCustomizationPlugin.java
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  /**
+   * Method to re-apply the command line customizations done during {@link DefaultPreferences#load} after modification
+   * of default values done from an other plugin.
+   * 
+   * @param preferencesNode
+   *      the default preference node to customize.
+   */
+  private void applyCommandLineCustomizations(IEclipsePreferences preferencesNode) {
```
sandupostaru: @lfasani If I understand correctly, we need to do this dirty hack, because we modify at runtime the default preferences, so we need to ensure that these changes are reflected in the command line default preferences.

What if instead we remove all of this code from here, and we put all the preferences in `org.polarsys.capella.rcp\plugin_customization.ini`, **then we don't need to do this hack anymore right?** The command line preferences will be automatically syncronized.

common/plugins/org.polarsys.capella.common.platform.sirius.customization/src/org/polarsys/capella/common/platform/sirius/customisation/SiriusCustomizationPlugin.java
```diff
@@ -0,0 +1,146 @@
...
+    defaultScope.putBoolean(SiriusUIPreferencesKeys.PREF_RELOAD_ON_LAST_EDITOR_CLOSE.name(), false);
+
+    // Set the default Sirius scale option (20%).
+    defaultScope.putInt(SiriusUIPreferencesKeys.PREF_SCALE_LEVEL_DIAGRAMS_ON_EXPORT.name(), 2);
+
+    // Don't use colors from odesign in diagram palettes
+    defaultScope.putBoolean(SiriusUIPreferencesKeys.PREF_DISPLAY_VSM_USER_FIXED_COLOR_IN_PALETTE.name(), false);
+
+    // Re-apply command line customizations
+    applyCommandLineCustomizations(defaultScope);
```
sandupostaru: I am talking about these preferences, see the comment below.

core/plugins/org.polarsys.capella.core.data.migration/src/org/polarsys/capella/core/data/migration/cmdline/MigrationCommandLine.java
```diff
@@ -104,15 +109,29 @@ public void migrateAllImportedProjects(Shell shell) {
    *
    */
   public void setRefreshPrefs() {
+      IPreferenceStore preferenceStore = SiriusEditPlugin.getPlugin().getPreferenceStore();
+      initialValue_RefreshOnOpening = preferenceStore.getBoolean(SiriusUIPreferencesKeys.PREF_REFRESH_ON_REPRESENTATION_OPENING.name());
```
sandupostaru: Is there a differences between accessing the preferences by using the `preferenceStore` like we here, compared to `InstanceScope.INSTANCE.getNode(SiriusEditPlugin.ID);` ? 

Because for obtaining the preferences we do it in the first way, and for resetting them we do it another way. I suppose the result is the same, but this is kinda confusing.

common/plugins/org.polarsys.capella.common.platform.sirius.customization/src/org/polarsys/capella/common/platform/sirius/customisation/SiriusCustomizationPlugin.java
```diff
@@ -0,0 +1,146 @@
...
+    defaultScope = DefaultScope.INSTANCE.getNode(SiriusEditPlugin.ID);
+    // Required since Sirius 5.1 since default behavior is not welcome
+    defaultScope.putBoolean(SiriusUIPreferencesKeys.PREF_SAVE_WHEN_NO_EDITOR.name(), false);
+    defaultScope.putBoolean(SiriusUIPreferencesKeys.PREF_RELOAD_ON_LAST_EDITOR_CLOSE.name(), false);
+
+    // Set the default Sirius scale option (20%).
+    defaultScope.putInt(SiriusUIPreferencesKeys.PREF_SCALE_LEVEL_DIAGRAMS_ON_EXPORT.name(), 2);
+
+    // Don't use colors from odesign in diagram palettes
+    defaultScope.putBoolean(SiriusUIPreferencesKeys.PREF_DISPLAY_VSM_USER_FIXED_COLOR_IN_PALETTE.name(), false);
```
sandupostaru: I am talking about these preferences, see the comment below.

common/plugins/org.polarsys.capella.common.platform.sirius.customization/src/org/polarsys/capella/common/platform/sirius/customisation/SiriusCustomizationPlugin.java
```diff
@@ -0,0 +1,146 @@
...
+    // Re-apply command line customizations.
+    if (preferencesNode instanceof DefaultPreferences) {
+      try {
+        ReflectionHelper.invokeMethod(preferencesNode, DefaultPreferences.class, "applyCommandLineDefaults", new Class[] {}, new Object[] {}, true); //$NON-NLS-1$
+      } catch (SecurityException | NoSuchMethodException | IllegalArgumentException | IllegalAccessException | InvocationTargetException e) {
+        getLog().log(new Status(IStatus.ERROR, PLUGIN_ID, IStatus.ERROR, "Failed to apply plugin customization from command line arguments for plugin " + PLUGIN_ID, e)); //$NON-NLS-1$
+      }
+    }
+  }
+}
```
sandupostaru: @lfasani If I understand correctly, we need to do this dirty hack, because we modify at runtime the default preferences, so we need to ensure that these changes are reflected in the command line default preferences.

What if instead we remove all of this code from here, and we put all the preferences in `org.polarsys.capella.rcp\plugin_customization.ini`, **then we don't need to do this hack anymore right?** The command line preferences will be automatically syncronized.

lfasani: After more tests it seems that using preferences.ini in the root of the capella plugin the default preference set by sirius are not overriden by Capella

In my opinion the way the default preference is initialized is
* PreferenceInitialiser Sirius
* overriden by the CommandLine 
* then overriden by PreferenceInitializer Capella (I would have thought it is the same thing with preference.ini but in my case preference.ini was not even taken into account)

Difficult to control what happen using preferences.ini so I prefer the way this commit has been done even if it looks like a hack.
In any case, the java doc indicates that preference should be initialized only by the plugin that declares it which is not what we do. So anyway, I do not do eat the right way.

PR #80, opened 2020-02-12
[559825] Add 3 command lines: Refresh Representations, Remove Hidden  Elements and Export All Images

jaubry-obeo:
Bug: 559825
Change-Id: I4fe9405b4cde0d8c9a3eeb4613eafa5bfef81405


pdulth:
Main idea;
- For refresh all, we need to run Capella workbench (CapellaWorkbenchAdvisor), but we have to wait until its fully started. To do that, we use a eclipse4 event, with org.polarsys.capella.core.commandline.core.ui.CommandLineWorkbenchModel. then, we call the commandLineApp. If we used an IStartup or workbench.advisor.postStartup, we are not fully launched.

- In cases where we don't want to open window, (validation rules) we use another WorkbenchAdvisor, and run the command line afterwards. (Without window, there is no eclipse4 event and no IStartups launched). We are not really in a proper workbench fully launched though, but like before.

- Jobs (Remove Hidden, Refresh All) weren't displaying any OK result when launched in Capella (only error or warning) but only in Error Log, not in Information View. 
In command line mode, logs were OK. 
=> I added a new property on Job (ALWAYS_LOG) to log any results in both mode, command line and in Capella, and in both views.

PR #89, opened 2020-02-14
V1.4.x rework on validation rules 53863

yugansan:


common/plugins/org.polarsys.capella.common.tools.report.appenders.reportlogview/src/org/polarsys/capella/common/tools/report/appenders/reportlogview/MarkerViewColumns.java
```diff
@@ -477,10 +481,17 @@ public Image getImage(Object element) {
             // resolution generators in plugins that aren't yet
             // loaded.
             IMarker marker = (IMarker) element;
+            Collection<IMarker> c = new HashSet<IMarker>();
+            c.add(marker);
             IMarkerResolution[] resolutions = IDE.getMarkerHelpRegistry().getResolutions(marker);
+            if (resolutions != null && resolutions.length > 0 && resolutions[0] instanceof ReportMarkerResolution) {
```
minhtutonthat: Wrong check. What if the first resolution in the list is not enabled and the second resolution is enabled?

PR #104, opened 2020-02-26
[560556] Drag and drop of deployed Behavior PC in other PC does not work anymore

jaubry-obeo:
Bug: 560556
Change-Id: Ie3754f856973e96f2720a5252d5f1cf27368bb13

### DIFF
--- a/tests/plugins/org.polarsys.capella.test.diagram.tools.ju/src/org/polarsys/capella/test/diagram/tools/ju/xab/DragAndDropPhysicalComponent.java
+++ b/tests/plugins/org.polarsys.capella.test.diagram.tools.ju/src/org/polarsys/capella/test/diagram/tools/ju/xab/DragAndDropPhysicalComponent.java
@@ -10,12 +10,29 @@
  *******************************************************************************/
 package org.polarsys.capella.test.diagram.tools.ju.xab;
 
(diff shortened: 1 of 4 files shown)


tests/plugins/org.polarsys.capella.test.diagram.tools.ju/src/org/polarsys/capella/test/diagram/tools/ju/xab/DragAndDropPhysicalComponent.java
```diff
@@ -25,18 +42,18 @@ public void test() throws Exception {
...
-    testOnXAB(context, PA__PAB_DIAGRAM, PA__PAB_DIAGRAM,
-        PA__PC_PART_PC14);
-    // drag and drop sub physical component
-    testOnXAB(context, PA__PAB_DIAGRAM, PA__PAB_COMPONENT_PC2,
-        PA__PC_PART_PC2_1);
+    oldTest(context);
+    newTest(session, context);
+  }
 
+  public void oldTest(SessionContext context) {
```
pdulth: does old tests are deprecated? shall be removed?
otherwise, can you rename these methods oldTests/newTests to something like "non deploy / deploy"

PR #108, opened 2020-03-03
[559825] fix NPE thrown when executing a commandline with"-forceoutputfoldercreation" argument but without "-outputfolder" argument

jaubry-obeo:
Change-Id: Iadacdf6baf2c5da7b9eb308c07fe149ffe383d89


pdulth:
Does this one is now deprecated with command line refactoring?

PR #110, opened 2020-03-03
[560715] Avoid inter-resource set dependencies with platform plug-in elements

oconstant:
Bug: 560715
Change-Id: I63676ea870a5e8bcc6969fd612701e9bcc6fae04
Signed-off-by: Olivier Constant <olivier.constant@thalesgroup.com>

### DIFF
--- a/core/plugins/org.polarsys.capella.core.compare/src/org/polarsys/capella/core/compare/CapellaScope.java
+++ b/core/plugins/org.polarsys.capella.core.compare/src/org/polarsys/capella/core/compare/CapellaScope.java
@@ -31,6 +31,7 @@ import org.eclipse.emf.ecore.EReference;
 import org.eclipse.emf.ecore.EcorePackage;
 import org.eclipse.emf.ecore.resource.Resource;
 import org.eclipse.emf.ecore.resource.ResourceSet;
+import org.eclipse.emf.ecore.util.EcoreUtil;
 import org.eclipse.emf.edit.domain.EditingDomain;
 import org.eclipse.emf.transaction.TransactionalEditingDomain;
 import org.eclipse.emf.transaction.util.TransactionUtil;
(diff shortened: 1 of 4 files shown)


core/plugins/org.polarsys.capella.core.compare/src/org/polarsys/capella/core/compare/CapellaScope.java
```diff
@@ -282,6 +292,36 @@ protected IFile getFileFor(Resource resource) {
...
+      Resource valueResource = element.eResource();
+      if (valueResource != null && localResourceSet != valueResource.getResourceSet()) {
+        // Element belongs to another resource set
+        URI valueResourceURI = valueResource.getURI();
+        if (valueResourceURI != null && valueResourceURI.isPlatformPlugin()) {
+          // And element belongs to a platform plug-in
+          URI valueURI = EcoreUtil.getURI(element);
+          if (valueURI != null) {
+            // Try and get the same element (URI-based) in the local resource set
+            EObject localElement = localResourceSet.getEObject(valueURI, false);
```
pdulth: Are you sure about the `false`?
Does external resources are loaded as long as scope is used in diffmerge?

oconstant: The rationale is to have as little impact as possible: if the desired element is not readily available, then the usual behavior is executed. It does not otherwise seem harmful to me because we are in a context (scope modification through add(...)) where the scope is supposed to have been loaded already. Also, the `FragmentedModelScope` base class was not intended to allow additional loading (using `true`) from a modification method such as `add(...)`.

pdulth: Ok, i have updated junit accordingly

sandupostaru:
[Added Junit](https://ci.eclipse.org/capella/job/capella-product/job/PR-110/3/testReport/org.polarsys.capella.test.platform.ju.testcases/DiffmergeExternalReferences/DiffmergeExternalReferences/) passed.

PR #126, opened 2020-03-10
Bugzilla/558030/search and replace

georgiana-ecobici:
Filter semantic elements.


core/plugins/org.polarsys.capella.core.ui.search/src/org/polarsys/capella/core/ui/search/searchfor/MetaClassesParticipantsItemProvider.java
```diff
@@ -90,16 +91,15 @@ public boolean isShowSemantics() {
   public void setShowSemantics(boolean showSemantics) {
     this.showSemantics = showSemantics;
   }
-
-  protected boolean isFiltered(Object cls) {
-    if(!showSemantics && !showAbstract)
-      return true;
+  
+  // based on abstract/semantic checks, check if the element is displayed or not
+  protected boolean isDisplayed(Object cls) {
```
minhtutonthat: This should take into account 2 filters at the same time. In this state, a semantic meta-class is displayed when the Semantic filter is checked regardless of the Abstract filter being checked or not. Similarly, an abstract class is displayed when the Abstract filter is checked regardless of the Semantic filter being checked or not.

core/plugins/org.polarsys.capella.core.ui.search/src/org/polarsys/capella/core/ui/search/searchfor/MetaClassesParticipantsItemProvider.java
```diff
@@ -90,16 +91,21 @@ public boolean isShowSemantics() {
...
-      return true;
+  
+  
+  // based on abstract and semantic checks, check if the element is displayed or not
+  protected boolean isDisplayed(Object cls) {
     if(cls instanceof EClass) {
-      EClass eCls = (EClass) cls;
-      if(showSemantics && !showAbstract && eCls.isAbstract()) {
+      EClass eclass = (EClass) cls;
+      boolean eclassIsSemantic = SimplifiedCapellaMetadata.INSTANCE.isSemantic(eclass);
```
minhtutonthat: The logic seems a bit confusing for me. In my opinion, you could do something like:
boolean abstractFilterPassed = showAbstract == eclass.isAbstract()
boolean semanticFilterPassed = showSemantics == implifiedCapellaMetadata.INSTANCE.isSemantic(eclass)
if (abstractFilterPassed && semanticFilterPassed) {
return true;
}
return false;

PR #129, opened 2020-03-10
[560956] Add IS2IS initialization

pdulth:


pdulth:
⚠️ To squash

core/plugins/org.polarsys.capella.core.projection.scenario/src/org/polarsys/capella/core/projection/scenario/topdown/TopDownExt.java
```diff
@@ -0,0 +1,399 @@
...
+            for (Role role : ((OperationalActivity) function).getAllocatingRoles()) {
+              componentBounds.addAll(role.getAllocatingEntities());
+            }
+          }
+
+        } else {
+          instances.add(function);
+        }
+      }
+    } else if (operation instanceof ExchangeItemAllocation) {
```
pdulth: To propagate SequenceMessage with ExchangeItemAllocation as operation, its here. We lookup for Components implementing/using the interface in the target Architecture

core/plugins/org.polarsys.capella.core.projection.scenario/src/org/polarsys/capella/core/projection/scenario/topdown/TopDownExt.java
```diff
@@ -0,0 +1,399 @@
...
+            }
+          }
+        }
+
+        if (!isLinked) {
+          partBounds.addAll(relatedParts);
+        }
+
+      } else {
+        ExchangeItemInstance instance = (ExchangeItemInstance) ScenarioExt.getExchangeItemInstanceRole(sourceEnd);
```
pdulth: To make ExchangeItem as instance role work, its here. 
targetOperation is the ExchangeItemAllocation to use (the one from targetArchitecture if it exist, or the original one if not transitioned according to preferences)
We lookup the original ExchangeItemInstance of the message end. If targetOperation is transitioned, then we lookup for the transitioned ExchangeItemInstance. Otherwise, we use the ExchangeItemInstance if it match with targetOperation

PR #139, opened 2020-03-16
Bugzilla/553820/v1.4.x

lredor:


sandupostaru:
@jaubry-obeo Can you take a look at the[ failing layout migration test](https://ci.eclipse.org/capella/job/capella-product/job/PR-139/1/testReport/junit/org.polarsys.capella.test.migration.ju.testcases.basic/SysmodelMigrationLayout/SysmodelMigrationLayout/) and fix it?

lredor:
Even in 1.4.0 it seems that the migration is KO. Not in automatic test but in real environment. When I migrate the data from 1.3.1 to 1.4.0 and open "[LFCD] Composite_FC - Logical Functional Chain Description" the result is KO. The compartment of "FC1" has an unexpected size:
image
Do you add "specific migration" in Capella to transfrom the diagrams of kind "Functional Chain Description" between 1.3.1 and 1.4.0?
There is maybe a change in 1.4.1 that breaks the layout, I will have a look. But even in 1.4.0 the result is wrong and can have side effects.

sandupostaru:
> 
> 
> Even in 1.4.0 it seems that the migration is KO. Not in automatic test but in real environment. When I migrate the data from 1.3.1 to 1.4.0 and open "[LFCD] Composite_FC - Logical Functional Chain Description" the result is KO. The compartment of "FC1" has an unexpected size:
> image
> Do you add "specific migration" in Capella to transfrom the diagrams of kind "Functional Chain Description" between 1.3.1 and 1.4.0?
> There is maybe a change in 1.4.1 that breaks the layout, I will have a look. But even in 1.4.0 the result is wrong and can have side effects.

Yes, we try to preserve the layout of old XFCD diagrams in 1.4.0.
Since previously Functional Chains could not be stacked (they could not contain children), this leads to the result you are seeing. 

**But this is know and expected! Preserving the old layout was more important**

jaubry-obeo:
In this previous commit https://github.com/eclipse/capella/commit/eb08731eca3778bdedc0675cc2e86fb934b80cef we have overwrite the "linux.layout" file because of infra change, but we have seen that some coordinates have been changed by a few pixels. Here, the SysmodelMigrationLayout testcase failed, but with the same pattern: 
- x:474 => x:475 => x:476
- x:503 => x:505 => x:507
- ...

Maybe that at each sirius migration, a position delta calculation is redone but should not. But we think is not a regression caused by "[553820] Add preferences for Jump links properties" development.

here the commit with coordinates change merged: https://github.com/eclipse/capella/pull/142/commits

sandupostaru:
Closed by #142

PR #146, opened 2020-03-18
[553820] Override default values for jump links properties

jaubry-obeo:
Cherry-pick from v1.4.x


pdulth:
https://github.com/eclipse/capella/pull/148 will have to be cherry picked too.

PR #149, opened 2020-03-19
Search in addon defaut setting implementation bugzilla/558030/search and replace

minhtutonthat:


core/plugins/org.polarsys.capella.core.ui.search/src/org/polarsys/capella/core/ui/search/searchfor/SearchForItemCache.java
```diff
@@ -109,28 +134,40 @@ public static SearchForItemCache getInstance() {
...
+
+  public Set<SearchForItem> getAttributeItems() {
+    return new HashSet<>(attributeName2AttributeItemMap.values());
+  }
+
+  public Set<SearchForItem> getAddonItems(Viewpoint vp) {
+    Set<String> vpClasses = classID2ViewpointMap.keySet().stream().filter(cls -> classID2ViewpointMap.get(cls) == vp)
+        .collect(Collectors.toSet());
+    return classID2ClassItemMap.keySet().stream().filter(vpClasses::contains)
+        .map(cls -> classID2ClassItemMap.get(cls)).collect(Collectors.toSet());
```
sandupostaru: Here it is faster to just iterate on the `vpClasses` instead of iterating on all of the `classID2ClassItemMap.keySet()`. And in this way `vpClasses` can just be a simple `List`, no need for a `Set`.

```java
vpClasses.stream().map(cls -> classID2ClassItemMap.get(cls)).filter(Objects::nonNull).collect(Collectors.toSet())
```

PR #154, opened 2020-03-20
552679 - Catch exceptions on session close with dirty state because a distant resource is not reachable

mPorhel:
See https://bugs.eclipse.org/bugs/show_bug.cgi?id=552679

This bug can be reproduced only with Team for Capella.


pdulth:
Are you sure you want to merge all of this in master?

mPorhel:
Hi Philippe, Sandu, 

It seems I targeted the wrong branch for this pull request, it was not for master but for v1.4.x.

It has been replaced by https://github.com/eclipse/capella/pull/160

Resulting commits: 
- v1.4.x: https://github.com/eclipse/capella/commit/262251439dadcce3c05c5999e4460e1e3b3c8f22
- master : https://github.com/eclipse/capella/commit/3a50bf91574bea052867b85fe54d036a67a0193b

PR #159, opened 2020-03-24
[560962] Check is show icon and is custom icon

pdulth:
Allow style customization on icon path.

Bug: 560962
Change-Id: If36c4fbe879da157916b5f38c3c2c230ca9aeffb
Signed-off-by: Florian Barbin <florian.barbin@obeo.fr>


pdulth:
@tguiu here

pdulth:
Why do we use a PartComponentStyleConfiguration instead of PartItemProvider ?

We use both. PartItemProvider is used to create part icon add foot balloon on Parts

Monopart (standard mode)
image

Multipart (ReusableComponents = true)
image

PR #166, opened 2020-03-25
Add the ability to extend search attribute items

minhtutonthat:
Change-Id: I6eb35c5bb567893418f7f2fd8457176b984c0d87


pdulth:
@minhtutonthat  @sandupostaru 
I think we need to wait for DOD and further refinement of search before trying to extend it in T4C.

For a first draft, I think SearchForAttribute is too restrictive as we need to lookup to references too, and more generally in queries. We have queries, editable or not, but attributes, references, are queries.

For me, 
- The search queries are the one available in mass visualization columns, 
- The search/replaceable queries are the one in mass edition.

That way, if we want to lookup for CDO id in T4C, then we will just have to add a query in semantic browser to make it available in search, but also in mass visualisation.
maybe we need to add a more general query extension point, not a search dedicated one.

PR #180, opened 2020-04-07
[releng][oomph] Setup model rewritten

felixdo:
* Simplify and adapt capella.setup for the following:
- Use github repositories
- Addons are now downloaded for all Capella Developers
- Target platform calculation much simpler now, only repo
  urls are needed
- Users of the current model should ditch their current installs
  and reinstall with the new setup

* Some test projects had missing natures
- This is important, because capella nature is used
  to identify test projects, which shouldn't be imported
  into the workspace

Signed-off-by: Felix Dorner <felix.dorner@gmail.com>

### DIFF
--- a/releng/plugins/org.polarsys.capella.settings/oomph/Capella.setup
+++ b/releng/plugins/org.polarsys.capella.settings/oomph/Capella.setup
@@ -16,748 +16,328 @@
     name="capella"
     label="Capella">
   <setupTask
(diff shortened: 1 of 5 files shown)


felixdo:
You dont have to test this .setup. After you merge it, I can use it to create a so-called configuration, which then simplifies/preselects available options in the setup. I want you to then test/try that Configuration. But I can only create it once the .setup is in the git repo.

PR #183, opened 2020-04-24
V1.4.x sirius6.3.1 s20200424 062347

lredor:


lredor:
Wrong target branch master instead of v1.4.x (pull request replaced by https://github.com/eclipse/capella/pull/184)

PR #158, opened 2020-03-24
561395 exception manage references

aliakar82:

PR #214, opened 2020-05-18
[561395] Exception is raised when trying to manage references for a model with broken reference to a library (#158)

minhtutonthat:
* [561395] Exception is raised when trying to manage references for a model
with broken reference to a library

Bug: 561395

Change-Id: I2ef8dbbcd7c6e3314bafe16c9bf55c6ba9cb165a
Signed-off-by: Ali AKAR <ali.akar82@gmail.com>


sandupostaru:
Can you please change the commit header 561395 exception manage references to match regular template. We forgot to do it on 1.4.x but we can at least fix it here.

PR #220, opened 2020-05-19
[563340] Clean code of SiriusSessionFactory.createSession

lredor:
This commit avoids to use the final parameter transactionalEditingDomain
that had the same name as the field transactionalEditingDomain of DASI.
This causes the creation, at compilation time, of a
val$transactionalEditingDomain in the DASI. This "new field" is not set
to null in DASI.close and can cause a memory leak.

Bug: 563340
Change-Id: I72d27ff022bbdc52534a7ff5b4fb95bf30665363


pdulth:
nice catch 👍

pdulth:
@lredor note that is code is duplicated in T4C.

PR #231, opened 2020-05-25
Doc - Enhancements of the implementation of Actors

MalinaStefaniaStoicanescu:
Signed-off-by: MalinaStoicanescu <stoicanescu.malina@gmail.com>


pdulth:
Hi Malina, 
Can you check which account mail do you use?

It seems that you have created eclipse.org account with your Thales mail whereas you have committed here with your gmail one. You need to have only one mail registered here and eclipse account.

In https://github.com/settings/emails , can you also uncheck "Keep my email addresses private" box.
Thanks

MalinaStefaniaStoicanescu:
> Hi Malina,
> Can you check which account mail do you use?
> 
> It seems that you have created eclipse.org account with your Thales mail whereas you have committed here with your gmail one. You need to have only one mail registered here and eclipse account.
> 
> In https://github.com/settings/emails , can you also uncheck "Keep my email addresses private" box.
> Thanks

Hi :) 
Thanks for your comment! The box is now unchecked. I will make another commit with the correct account email.

PR #232, opened 2020-05-25
[558032] Insert title Block and Capella feature values inside Diagram

pdulth:


pdulth:
for reminder, Doc Capella Guide > Diagrams still contains both tools, 
`Diagram Title Block` / `Element Title Block`

PR #244, opened 2020-05-28
Diagram Title Block Preferences fixes

pdulth:
## before
image

## after
image

On Move over
image


pdulth:
Untitled-1
Untitled-3

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/preferences/TitleBlockPreferencePage.java
```diff
@@ -255,19 +255,12 @@ public boolean performOk() {
...
           }
           currentTableContent = currentTableContent + tccMatrix.get(i).get(j).name + SEPARATOR
               + tccMatrix.get(i).get(j).content;
-        } else {
-          if (tccMatrix.get(i).get(j).name.equals(EMPTY_STRING)
-              || tccMatrix.get(i).get(j).content.equals(EMPTY_STRING)) {
-            createMessageBox(EDIT_ALL_CELLS, ERROR);
-            return false;
-          }
+        } else {
```
sandupostaru: The user can add a line or a column at any position, not only at the end.
Why do we only check the last cell ?

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/preferences/TitleBlockPreferencePage.java
```diff
@@ -255,19 +255,12 @@ public boolean performOk() {
...
       for (int j = 0; j < columnsNumber; j++) {
         if (i == linesNumber - 1 && j == columnsNumber - 1) {
-          if (tccMatrix.get(i).get(j).name.equals(EMPTY_STRING)
-              || tccMatrix.get(i).get(j).content.equals(EMPTY_STRING)) {
-            createMessageBox(EDIT_ALL_CELLS, ERROR);
-            return false;
+          if (tccMatrix.get(i).get(j).content.equals(EMPTY_STRING)) {
+        	  tccMatrix.get(i).get(j).content = " ";
           }
           currentTableContent = currentTableContent + tccMatrix.get(i).get(j).name + SEPARATOR
```
sandupostaru: Please use a StringBuilder for the `currentTableContent`.

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/preferences/TitleBlockPreferencePage.java
```diff
@@ -66,7 +69,7 @@ public TitleBlockCell(String name, String content) {
 
     @Override
     public String toString() {
-      return name;
+      return name+"\n"+content; //$NON-NLS-1$
```
sandupostaru: Slight formatting issue ;) 
There are others on this file so a global format should do the job.

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/preferences/TitleBlockPreferenceField.java
```diff
@@ -208,24 +201,21 @@ protected void doLoadDefault() {
...
   @Override
   protected void doStore() {
-    String currentTableContent = EMPTY_STRING;
+    StringBuilder table = new StringBuilder();
+    
     for (int i = 0; i < linesNumber; i++) {
       for (int j = 0; j < columnsNumber; j++) {
-        if (i == linesNumber - 1 && j == columnsNumber - 1) {
-          if (tccMatrix.get(i).get(j).content.equals(EMPTY_STRING)) {
-            tccMatrix.get(i).get(j).content = " ";
```
pdulth: fixed by https://github.com/eclipse/capella/pull/244/commits/4f40ea4c31f5a8c3fa5c6781922be87f240bf43a#diff-24254bdb436064c4b318f9508990f052R269

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/preferences/TitleBlockPreferencesInitializer.java
```diff
@@ -14,73 +14,53 @@
...
-  }
-  
-  
-  /**
-   * @return IPreferenceStore the default preference store instance
-   */
-  public static IPreferenceStore doGetPreferenceStore() {
-    return Activator.getDefault().getPreferenceStore();
+    IPreferenceStore preferenceStore = ScopedCapellaPreferencesStore.getInstance(CapellaActionsActivator.PLUGIN_ID);
+    preferenceStore.setDefault(TitleBlockPreferencePage.COLUMNS_NUMBER_PREFERENCE_STORE, 2);
```
pdulth, line 29: Just for reminder:
org.polarsys.capella.core.commands.preferences.service.AbstractPreferencesInitializer are inconsistent and not working properly.

`putInt` If the InstanceScope contains a value, at restart the default value will contains the previous instance value, not the one given in parameter. It will never set only the default value. always in instance scope.

`putBoolean` and `putString` used in https://github.com/eclipse/capella/blob/6ef8e899f3503c655a1c528a2d33442dc722e90d/core/plugins/org.polarsys.capella.core.model.preferences/src/org/polarsys/capella/core/model/preferences/DeletePreferences.java#L35 do nothing if DefaultScope.

If ProjectScope, like used in initializer, as there is never a selected project at startup, its also weird.

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/preferences/TitleBlockPreferencePage.java
MalinaStefaniaStoicanescu: If the last cell contains an empty content, when string is build, it will appear something like this: String s = "...SEPARATORSEPARATOR". When perform split on this string, it will create an array that will not contain the empty content of the last cell (and instead of 12 elements for example, it will contain only 11). Mostly for this is that check. Maybe it could be removed that else and add SEPARATOR after every content/name, even if is the last cell, I'm not sure about this :-?

PR #246, opened 2020-06-01
Fix Deletion of Title Block Elements

sandupostaru:
- Deletion of a Title Block cell clears the cell content
- Deletion of a Title Block deletes elements as expected, except for any
referenced Model Elements

Change-Id: If8a26cbcacab8562aa1d7752dfc5ec13d9480bfa
Signed-off-by: Sandu Postaru <sandu.postaru@thalesgroup.com>

### DIFF
--- a/core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/delete/CapellaDeleteActionHook.java
+++ b/core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/delete/CapellaDeleteActionHook.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2006, 2017 THALES GLOBAL SERVICES.
+ * Copyright (c) 2006, 2017, 2020 THALES GLOBAL SERVICES.
  * 
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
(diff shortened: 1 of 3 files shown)


sandupostaru:
The deletion of the Referenced Model Elements in Titlle Blocks can be done in two different places.

- At the **Capella Level** in `CapellaDeleteActionHook.beforeDeleteCommandExecution` where we extract the target elements for the selected diagram elements. The fix here is simple, ignore the semantic elements that are the target of a diagram element that has a Title Block mapping.

- At the **Sirius Level**. According to the [Sirius Documentation](https://www.eclipse.org/sirius/doc/specifier/diagrams/Diagrams.html#delete_too) If you do not associate a Delete tool to a mapping, the default delete behavior when the user deletes a graphical element is to remove the target semantic element 
and all the associated semantic elements from the semantic model. 

Since we want a different behavior we define this explicit Delete tool that does not delete the model element.
When the Parent Mapping is Deleted, the delete function is also invoked on the children (in our case we do nothing at the semantic level, so the model element is not deleted) and then the children diagram node is removed by Sirius.

If we add a precondition that is always false thus disabling the delete, then the deletion of the parent will not succeed, since some of its children can no be deleted ...

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/delete/CapellaDeleteActionHook.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2006, 2017 THALES GLOBAL SERVICES.
+ * Copyright (c) 2006, 2017, 2020 THALES GLOBAL SERVICES.
```
pdulth, line 2: It shall be "2006, 2020"
But don't bother with updating them manually.

sandupostaru: I will wait for the build to be done and if successful I will fix it in another commit and we can merge right after.

PR #289, opened 2020-06-29
564740 v1.4.x bug show hide component

minhtutonthat:


minhtutonthat:
For the first bug, the fix is to add an "If task" before the "Delete task" (like what we did in PAB for PC). If the If task wasn't there and we try to delete a view which is already removed, the context would become the semantic element and the Sirius warning would appear in the Error log.
For the second bug, the tool just didn't work. It was coded again to be similar to other Show/Hide component tools.

minhtutonthat:
There're still failed tests to fix

PR #303, opened 2020-07-09
M2 Annotations for XMLPiviot/Orchestra Connector

felixdo:
So these 2 affect xml pivot and also the connector. It's not required urgently, but we should consider this before 5.0, as I had to hardcode some hacks into xmlpivot as workarounds.

I didn't generate java code though... should I, or can you do it?


sandupostaru:
Closed by #395

PR #308, opened 2020-07-22
Fix compilation problems with TP 2020-06

lredor:


sandupostaru:
Fixed by https://github.com/eclipse/capella/pull/310.
Thanks for the heads up!

PR #311, opened 2020-07-30
[test] Avoid the installation of the test DAnalysisSelector

lfasani:
The org.polarsys.capella.test.diagram.common.ju plugin declares a
TestSessionManagerListener which will inject a test DAnalysisSelector in 
the Sirius Session.
It is sometimes painful when working in a development context because it
changes the runtime behavior and breaks a lot of tests.

This commit allows to deactivate the TestSessionManagerListener using
the system property "capella.tests.disablecustomdanalysisselector".

Change-Id: I07290b600abfa0836c8c21ec4b2faa3be987d5b7

### DIFF
--- a/tests/plugins/org.polarsys.capella.test.diagram.common.ju/src/org/polarsys/capella/test/diagram/common/ju/headless/selector/TestSessionManagerListener.java
+++ b/tests/plugins/org.polarsys.capella.test.diagram.common.ju/src/org/polarsys/capella/test/diagram/common/ju/headless/selector/TestSessionManagerListener.java
@@ -20,12 +20,14 @@ import org.eclipse.ui.PlatformUI;
 
 public class TestSessionManagerListener extends SessionManagerListener.Stub {
 
-  @Override
-  public void notify(Session updated, int notification) {
-    super.notify(updated, notification);
-    boolean enableHeadless = (PlatformUI.getTestableObject().getTestHarness() != null);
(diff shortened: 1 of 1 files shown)


pdulth:
Hi, 
Can you describe when this problem occurs?
Is it when you run sirius junits in debug while this plugin is opened in your workspace?

lfasani:
in T4C dev environment, we use t4c_allDependencies.target that includes org.polarsys.capella.test.diagram.common.ju plugin.
So each time we start a test, we need to edit its launch config and remove the plugin. (with Windows it takes one minute for each launch config because displaying launch config is very very long) It is painful and a lackof time. Sometimes some people do not think about that at all and loose time trying to understand why the test is KO.

pdulth:
OK, my concern is about the system property `capella.tests.disablecustomdanalysisselector`, 
how your developers would know that this property exists?
registering the selector only on some tests using it seems easier no?

lfasani:
>how your developers would know that this property exists? 
This property system would be set in our abstracttest class so the develpper do not have to know it

>registering the selector only on some tests using it seems easier no?
As I said in my initial email, YES it would be a better solution. Easier? i don't know but better, yes. I did not do it because I was not sure of how to do it so that every Capella test that need it get it effectively.

pdulth:
Closed, this will be fixed by https://github.com/eclipse/capella/pull/331

PR #312, opened 2020-07-30
Fix failing tests 2020 06

sandupostaru:


sandupostaru:
Based on https://github.com/eclipse/capella/pull/310, so merge only after that PR is merged ;)
This PR is isolated so that work can be continued here for the remaining of the failing tests.

PR #328, opened 2020-08-21
[566264] Some operations might create unwanted Entities as a side effect

sandupostaru:
The problem resides in the use of the
`org.polarsys.capella.core.model.helpers.BlockArchitectureExt.getFirstComponent(BlockArchitecture,
boolean: TRUE)` method in functions called at the Operational Level.

Since there is no Operational System concept at that level, this method
constantly creates a new Entity at each call.

Change-Id: I3f031760613b3c6759910bce56ea1d479d7a9e85
Signed-off-by: Sandu Postaru <sandu.postaru@thalesgroup.com>

### DIFF
--- a/core/plugins/org.polarsys.capella.core.model.helpers/src/org/polarsys/capella/core/model/helpers/BlockArchitectureExt.java
+++ b/core/plugins/org.polarsys.capella.core.model.helpers/src/org/polarsys/capella/core/model/helpers/BlockArchitectureExt.java
@@ -684,15 +684,24 @@ public class BlockArchitectureExt {
    * @param architecture
    * @param create
    * @return
+   * 
+   * @deprecated use {@link BlockArchitecture::getSystem} or {@link BlockArchitectureExt::getOrCreateComponent} instead.
    */
+  @Deprecated
   public static Component getFirstComponent(BlockArchitecture architecture, boolean create) {
     Component first = null;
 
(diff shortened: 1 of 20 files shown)


core/plugins/org.polarsys.capella.core.model.helpers/src/org/polarsys/capella/core/model/helpers/BlockArchitectureExt.java
```diff
@@ -684,15 +684,24 @@ public static boolean isDefaultNameRootFunction(AbstractNamedElement element) {
...
+      EList<Entity> ownedEntities = entityPkg.getOwnedEntities();
+
+      first = ownedEntities.stream().filter(x -> !x.isActor()).findFirst().orElse(null);
+
       if ((first == null) && create) {
         first = OaFactory.eINSTANCE.createEntity(NamingConstants.CreateOaAnalysisCmd_entity_name);
-        ((EntityPkg) getComponentPkg(architecture, true)).getOwnedEntities().add((Entity) first);
+        ownedEntities.add((Entity) first);
+
+        CapellaElementExt.creationService(first);
```
sandupostaru, line 704: > 
> 
> the getOrCreateSystem can still be dangerous on oa, as it will create always a new entity.
> 
> maybe we can replace oa.getSystem call by a getFirstEntity?

Done.

pdulth (approved):

PR #329, opened 2020-08-21
[releng] Update Tycho 2.0 and use openjdk-jdk14-latest

pdulth:


pdulth:
I will do a rebase from https://github.com/eclipse/capella/pull/352 as javax.xml.bind is no longer required in manifest

sandupostaru (approved):

PR #339, opened 2020-08-27
[566434] Always call SemanticBrowserView.setInput in UI Thread

lredor:
This avoids potential "Invalid thread access".

Bug: 566434
Change-Id: I842eabddab9d3cae38ef7a5f43abbf020c351541

### DIFF
--- a/core/plugins/org.polarsys.capella.core.ui.semantic.browser/src/org/polarsys/capella/core/ui/semantic/browser/view/SemanticBrowserView.java
+++ b/core/plugins/org.polarsys.capella.core.ui.semantic.browser/src/org/polarsys/capella/core/ui/semantic/browser/view/SemanticBrowserView.java
@@ -120,11 +120,19 @@ import org.polarsys.capella.core.ui.semantic.browser.model.SemanticBrowserModel;
  */
 public abstract class SemanticBrowserView extends ViewPart implements ISemanticBrowserViewPart,
     ITabbedPropertySheetPageContributor, IEditingDomainProvider, IReadOnlyListener {
-
+  
   /**
    * Listener that listens to closing and closed session events.
    */
   protected class CloseSessionListener extends SessionManagerListener.Stub {
+    
+    Runnable cleaner = new Runnable() {
+      @Override
+      public void run() {
+        clean();
+      }
+    };
+    
     @Override
     public void notify(final Session updated, final int notification) {
       switch (notification) {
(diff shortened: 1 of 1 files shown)


core/plugins/org.polarsys.capella.core.ui.semantic.browser/src/org/polarsys/capella/core/ui/semantic/browser/view/SemanticBrowserView.java
```diff
@@ -397,7 +399,20 @@ public void clean() {
     // No need to set focus.
     boolean restoreState = shouldSetFocus ? true : false;
     shouldSetFocus = false;
-    setInput(null);
+
+    // Ensure execution of setInput in UI thread.
+    if (Display.getCurrent() == null) {
+        EclipseUIUtil.displayAsyncExec(new Runnable() {
```
SteveMonnier, line 405: I think is shouldn't be executed in async in order to have the same process as before but using the UI Thread

SteveMonnier (changes requested):

core/plugins/org.polarsys.capella.core.ui.semantic.browser/src/org/polarsys/capella/core/ui/semantic/browser/view/SemanticBrowserView.java
pdulth: @lredor any thought about this?

lredor: Oups, I missed the Steve comment. So I use async because there is no real reason to change the input in sync mode. If this method is called not in the UI thread, the UI thread is probably always in progress. Async allows to to the input refresh "later".

pdulth:
I have pushed another commit inspired from DDiagramEditorImpl
https://github.com/pcdavid/org.eclipse.sirius/blob/a88f229dca2559ce39764e9db5e3ce54fa211ae2/plugins/org.eclipse.sirius.diagram.ui/src-diag/org/eclipse/sirius/diagram/ui/tools/internal/editor/DDiagramEditorImpl.java#L1384-L1388

Now, all content of the `clean` method is done in the same thread, the 
https://github.com/eclipse/capella/pull/339/commits/28992b1115af746498f26152fac43f3404ab591c#diff-d8f673eca083b1ceb52f845189284ad46dd940aac718b7d3064acdba5523b4caR416-R418
could have be done before the async execution.

minhtutonthat (approved):

PR #341, opened 2020-08-27
[566437] Use IllegalStateExceptionQuery where it is necessary

lredor:
These cases have been detected using Team For Capella after a shutdown
of the server with a session opened.

Bug: 566437
Change-Id: I9a4a3e292a0024efb9fdf7f13e1ce2c030003e0b


pdulth (changes requested):
can you look at your modified files, it seems that tab are used instead of spaces in the catch part.
otherwise looks ok

lredor:
> can you look at your modified files, it seems that tab are used instead of spaces in the catch part.
> otherwise looks ok

I will look (the new change does not correspond to your remark but to a new discover case)

pdulth (approved):

PR #366, opened 2020-09-13
Master 566828 rcptt

aliakar82:


pdulth:
Created as junit instead 
https://github.com/eclipse/capella/pull/453

PR #381, opened 2020-10-01
[563193] Capella Diagram Performance issues

nlepine:
Signed-off-by: Nathalie Lepine <nathalie.lepine@obeo.fr>

### DIFF
--- a/core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sequencediag/ScenarioService.java
+++ b/core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sequencediag/ScenarioService.java
@@ -18,7 +18,6 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
-import java.util.Stack;
 
 import org.eclipse.emf.ecore.EObject;
 import org.eclipse.emf.ecore.util.EcoreUtil;
@@ -66,7 +65,6 @@ import org.polarsys.capella.core.data.interaction.InstanceRole;
 import org.polarsys.capella.core.data.interaction.InteractionFragment;
 import org.polarsys.capella.core.data.interaction.InteractionOperand;
 import org.polarsys.capella.core.data.interaction.InteractionPackage;
-import org.polarsys.capella.core.data.interaction.InteractionState;
 import org.polarsys.capella.core.data.interaction.MessageEnd;
 import org.polarsys.capella.core.data.interaction.MessageKind;
 import org.polarsys.capella.core.data.interaction.Scenario;
@@ -91,6 +89,7 @@ import org.polarsys.capella.core.sirius.analysis.CapellaServices;
 import org.polarsys.capella.core.sirius.analysis.IMappingNameConstants;
 import org.polarsys.capella.core.sirius.analysis.InformationServices;
 import org.polarsys.capella.core.sirius.analysis.SequenceDiagramServices;
+import org.polarsys.capella.core.sirius.analysis.cache.ScenarioCache;
 
 import com.google.common.base.Predicate;
 import com.google.common.base.Predicates;
@@ -884,95 +883,13 @@ public class ScenarioService {
    * @param ir
    * @return
    */
-  public List<StateFragment> getInteractionStatesOnExecution(InstanceRole ir) {
-    List<StateFragment> result = new ArrayList<StateFragment>(1);
-    List<InteractionFragment> fragments = SequenceDiagramServices
-        .getOrderedInteractionFragments((Scenario) ir.eContainer());
-    Stack<TimeLapse> execStack = new Stack<TimeLapse>();
-
-    for (InteractionFragment ifg : fragments) {
-      if (ifg.getCoveredInstanceRoles().contains(ir)) {
-        if ((ifg instanceof InteractionState) && execStack.isEmpty()) {
-          result.add((StateFragment) getStartingExecution(ifg));
-        }
-
-        TimeLapse startingExec = getStartingExecution(ifg);
-        if ((startingExec != null) && (startingExec instanceof Execution)) {
-          execStack.push(startingExec);
-        }
-
-        TimeLapse endingExec = getEndingExecution(ifg);
-        if ((endingExec != null) && (endingExec instanceof Execution) && !(execStack.isEmpty())) {
-          execStack.pop();
-        }
-      }
-    }
-    return result;
+  public Collection<StateFragment> getInteractionStatesOnExecution(InstanceRole ir) {
+    return ScenarioCache.getInstance().getStateFragmentSemanticCandidates(ir, ir);
   }
 
-  public List<StateFragment> getInteractionStatesOnExecution(Execution exec) {
-    List<StateFragment> result = new ArrayList<StateFragment>(1);
-    InstanceRole ir = exec.getCovered();
-
-    List<InteractionFragment> fragments = SequenceDiagramServices
-        .getOrderedInteractionFragments((Scenario) exec.eContainer());
-    Stack<TimeLapse> execStack = new Stack<TimeLapse>();
-    boolean inCurrentExec = false;
-    for (InteractionFragment ifg : fragments) {
-      if (ifg.getCoveredInstanceRoles().contains(ir)) {
-        if (exec.getStart() == ifg) {
-          inCurrentExec = true;
-        }
-
-        if (inCurrentExec && (ifg instanceof InteractionState) && !(execStack.isEmpty())
-            && (execStack.peek() == exec)) {
-          result.add((StateFragment) getStartingExecution(ifg));
-        }
-
-        TimeLapse startingExec = getStartingExecution(ifg);
-        if ((startingExec != null) && !(startingExec instanceof CombinedFragment)) {
-          execStack.push(startingExec);
-        }
-
-        TimeLapse endingExec = getEndingExecution(ifg);
-        if ((endingExec != null) && !(endingExec instanceof CombinedFragment) && !(execStack.isEmpty())) {
-          execStack.pop();
-        }
-
-        if (exec.getFinish() == ifg) {
-          inCurrentExec = false;
-        }
-      }
-    }
-    return result;
-  }
-
-  /**
-   * @param ifg
-   * @return
-   */
-  private TimeLapse getEndingExecution(InteractionFragment ifg) {
-    Scenario s = SequenceDiagramServices.getScenario(ifg);
-    for (TimeLapse lap : s.getOwnedTimeLapses()) {
-      if (lap.getFinish() == ifg) {
-        return lap;
-      }
-    }
-    return null;
-  }
-
-  /**
-   * @param ifg
-   * @return
-   */
-  private TimeLapse getStartingExecution(InteractionFragment ifg) {
-    Scenario s = SequenceDiagramServices.getScenario(ifg);
-    for (TimeLapse lap : s.getOwnedTimeLapses()) {
-      if (lap.getStart() == ifg) {
-        return lap;
-      }
-    }
-    return null;
+  public Collection<StateFragment> getInteractionStatesOnExecution(Execution exec) {
+    return ScenarioCache.getInstance().getStateFragmentSemanticCandidates(SequenceDiagramServices.currentInstanceRole(exec),
+        exec);
   }
 
   public List<InstanceRole> getCoveredFromAbstractFragment(AbstractFragment af) {
--- a/core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/SequenceDiagramServices.java
+++ b/core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/SequenceDiagramServices.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2006, 2018 THALES GLOBAL SERVICES.
+ * Copyright (c) 2006, 2020 THALES GLOBAL SERVICES.
  * 
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -16,6 +16,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 
 import org.eclipse.core.runtime.IStatus;
@@ -52,6 +53,7 @@ import org.polarsys.capella.core.data.interaction.MessageKind;
 import org.polarsys.capella.core.data.interaction.Scenario;
 import org.polarsys.capella.core.data.interaction.ScenarioKind;
 import org.polarsys.capella.core.data.interaction.SequenceMessage;
+import org.polarsys.capella.core.data.interaction.StateFragment;
 import org.polarsys.capella.core.data.interaction.TimeLapse;
 import org.polarsys.capella.core.data.oa.OperationalActivity;
 import org.polarsys.capella.core.data.oa.OperationalCapability;
@@ -59,32 +61,35 @@ import org.polarsys.capella.core.model.helpers.ScenarioExt;
 import org.polarsys.capella.core.model.helpers.SequenceMessageExt;
 import org.polarsys.capella.core.platform.sirius.ui.commands.CapellaDeleteCommand;
 import org.polarsys.capella.core.sirius.analysis.activator.SiriusViewActivator;
+import org.polarsys.capella.core.sirius.analysis.cache.ScenarioCache;
+import org.polarsys.capella.core.sirius.analysis.cache.ScenarioCache.SemanticCandidateContext;
 
 public class SequenceDiagramServices {
 
-  public static List<Execution> getExecutionsFromInstanceRoleOrParentExecution(EObject obj) {
-    if (obj instanceof InstanceRole) {
-      return getExecutionsFromInstanceRole((InstanceRole) obj);
-    }
-    return getExecutionFromExecution((Execution) obj);
-  }
-
-  private static Execution top(List<Execution> stack) {
-    if (stack.size() == 0) {
+  /**
+   * Return covered instance role for EObject.
+   * 
+   * @param eObject
+   *          EObject
+   * @return covered instance role.
+   */
+  public static InstanceRole currentInstanceRole(EObject eObject) {
+    if (eObject instanceof InstanceRole) {
+      return (InstanceRole) eObject;
+    } else if (eObject instanceof Execution) {
+      return ScenarioCache.getInstance().getInteractionCache(Execution::getCovered, (Execution) eObject);
+    } else if (eObject instanceof AbstractEnd) {
+      return ScenarioCache.getInstance().getInteractionCache(AbstractEnd::getCovered, (AbstractEnd) eObject);
+    } else if (eObject instanceof StateFragment) {
+      return ScenarioCache.getInstance()
+          .getInteractionCache(state -> currentInstanceRole(((StateFragment) state).getStart()), eObject);
+    } else {
       return null;
     }
-    return (stack.get(stack.size() - 1));
   }
 
-  /**
-   * Return executions which are just under an other execution.
-   * 
-   * @param objP
-   *          the execution
-   * @return the list of result executions
-   */
-  private static List<Execution> getExecutionFromExecution(Execution execution) {
-    return SequenceMessageExt.getExecutionFromExecution(execution);
+  public static Collection<Execution> getExecutionsFromInstanceRoleOrParentExecution(EObject eObject) {
+    return ScenarioCache.getInstance().getExecutionSemanticCandidates(currentInstanceRole(eObject), eObject);
   }
 
   public static List<InteractionFragment> getOrderedInteractionFragments(Scenario scenario) {
@@ -92,40 +97,6 @@ public class SequenceDiagramServices {
       return new ArrayList<InteractionFragment>();
     }
     return scenario.getOwnedInteractionFragments();
-  }
-
-  private static List<Execution> getExecutionsFromInstanceRole(InstanceRole ir) {
-    List<Execution> result = new ArrayList<Execution>();
-
-    Scenario scenario = (Scenario) ir.eContainer();
-    List<Execution> executionStack = new ArrayList<Execution>();
-    for (InteractionFragment ifgt : getOrderedInteractionFragments(scenario)) {
-      if (ifgt instanceof AbstractEnd) {
-        AbstractEnd ae = (AbstractEnd) ifgt;
-        if (ae.getCovered() == ir) {
-          // if ae starts an exception, the execution is stacked
-          // if ae ends an exception, the execution is unstacked
-          for (TimeLapse laptime : scenario.getOwnedTimeLapses()) {
-            if (laptime instanceof Execution) {
-              Execution exec2 = (Execution) laptime;
-              if (exec2.getCovered() == ir) {
-                if (exec2.getStart() == ae) {
-                  if (top(executionStack) == null) {
-                    result.add(exec2);
-                  }
-                  executionStack.add(exec2);
-                }
-                if (exec2.getFinish() == ae) {
-                  executionStack.remove(exec2);
-                }
-              }
-            }
-          }
-        }
-      }
-    }
-
-    return result;
   }
 
   public static EObject getSendingEndEvent(SequenceMessage message) {
@@ -141,16 +112,13 @@ public class SequenceDiagramServices {
     if (end == null) {
       return message; // found message case
     }
+    InstanceRole currentInstanceRole = currentInstanceRole(end);
+    List<SemanticCandidateContext> semanticCandidateContexts = ScenarioCache.getInstance().getSemanticCandidateContexts(currentInstanceRole);
 
-    Scenario sc = (Scenario) message.eContainer();
-    for (TimeLapse exec : sc.getOwnedTimeLapses()) {
-      if (exec.getFinish() == end) {
-        return exec;
-      }
-    }
-    Execution execution = getExecutionOfMessageEnd(end);
+    Optional<SemanticCandidateContext> sendingEvent = semanticCandidateContexts.stream()
+        .filter(ec -> ec.isStart() && ec.getElement().equals(message)).findFirst();
+    return sendingEvent.isPresent() ? sendingEvent.get().getParent() : currentInstanceRole;
 
-    return execution == null ? end.getCovered() : execution;
   }
 
   public static EObject getReceivingEnd(SequenceMessage message) {
@@ -158,52 +126,13 @@ public class SequenceDiagramServices {
     if (end == null) {
       return message; // lost message case
     }
-    Scenario sc = (Scenario) message.eContainer();
-    for (TimeLapse exec : sc.getOwnedTimeLapses()) {
-      if (exec.getStart() == end) {
-        return exec;
-      }
-    }
-    Execution execution = getExecutionOfMessageEnd(end);
-    return execution == null ? end.getCovered() : execution;
-  }
+    InstanceRole currentInstanceRole = currentInstanceRole(end);
+    List<SemanticCandidateContext> semanticCandidateContexts = ScenarioCache.getInstance().getSemanticCandidateContexts(currentInstanceRole);
 
-  /**
-   * return the execution which start the messageEnd taken in parameter.
-   * 
-   * @param end
-   * @return
-   */
-  private static Execution getExecutionOfMessageEnd(MessageEnd end) {
-    Scenario scenario = getScenario(end);
-    InstanceRole currentIR = end.getCovered();
-    List<Execution> executionStack = new ArrayList<Execution>();
-    for (InteractionFragment ifgt : scenario.getOwnedInteractionFragments()) {
-      if (ifgt instanceof AbstractEnd) {
-        AbstractEnd ae = (AbstractEnd) ifgt;
-        if (ae.getCovered() == currentIR) {
-          // if ae starts an exception, the execution is stacked
-          // if ae ends an exception, the execution is unstacked
-          for (TimeLapse laptime : scenario.getOwnedTimeLapses()) {
-            if (laptime instanceof Execution) {
-              Execution execution = (Execution) laptime;
-              if (execution.getCovered() == currentIR) {
-                if (execution.getStart() == ae) {
-                  executionStack.add(execution);
-                }
-                if (execution.getFinish() == ae) {
-                  executionStack.remove(execution);
-                }
-              }
-            }
-          }
-          if (ae == end) {
-            return top(executionStack);
-          }
-        }
-      }
-    }
-    return null;
+    Optional<SemanticCandidateContext> receivingEvent = semanticCandidateContexts.stream()
+        .filter(ec -> !ec.isStart() && ec.getElement().equals(message)).findFirst();
+    return receivingEvent.isPresent() ? receivingEvent.get().getParent() : currentInstanceRole;
+
   }
 
   /**
@@ -247,8 +176,8 @@ public class SequenceDiagramServices {
     if (object instanceof SequenceMessage) {
       SequenceMessage reorderedMessage = (SequenceMessage) object;
       reorderSequenceMessage(scenario, newPredecessor, newPredecessorOfEnd, reorderedMessage);
-    }  else if (object instanceof TimeLapse) {
-        TimeLapse execution = (TimeLapse) object;
+    } else if (object instanceof TimeLapse) {
+      TimeLapse execution = (TimeLapse) object;
       reorderTimeLapse(scenario, newPredecessor, newPredecessorOfEnd, execution);
     } else if (object instanceof InteractionState) {
       InteractionState interactionState = (InteractionState) object;
@@ -256,7 +185,7 @@ public class SequenceDiagramServices {
     } else if (object instanceof InteractionOperand) {
       InteractionOperand interactionOperand = (InteractionOperand) object;
       reorderInteractionOperand(scenario, newPredecessor, newPredecessorOfEnd, interactionOperand);
-    } 
+    }
     return object;
   }
 
@@ -276,16 +205,16 @@ public class SequenceDiagramServices {
     int currentFragmentIndex = -1;
     if (currentFragment != null) {
       currentFragmentIndex = scenario.getOwnedInteractionFragments().indexOf(currentFragment);
-    int currentIndexNewPredecessor = -1;
-    if (newPredecessor != null) {
-      currentIndexNewPredecessor = scenario.getOwnedInteractionFragments().indexOf(newPredecessor);
-    }
+      int currentIndexNewPredecessor = -1;
+      if (newPredecessor != null) {
+        currentIndexNewPredecessor = scenario.getOwnedInteractionFragments().indexOf(newPredecessor);
+      }
 
-    // Reorder shall use move method instead of remove/add
-    if (currentFragmentIndex < currentIndexNewPredecessor) {
-      scenario.getOwnedInteractionFragments().move(currentIndexNewPredecessor, currentFragmentIndex);
-    } else if (currentFragmentIndex > currentIndexNewPredecessor) {
-      scenario.getOwnedInteractionFragments().move(currentIndexNewPredecessor + 1, currentFragmentIndex);
+      // Reorder shall use move method instead of remove/add
+      if (currentFragmentIndex < currentIndexNewPredecessor) {
+        scenario.getOwnedInteractionFragments().move(currentIndexNewPredecessor, currentFragmentIndex);
+      } else if (currentFragmentIndex > currentIndexNewPredecessor) {
+        scenario.getOwnedInteractionFragments().move(currentIndexNewPredecessor + 1, currentFragmentIndex);
       }
     }
   }
@@ -300,9 +229,9 @@ public class SequenceDiagramServices {
       InteractionFragment newPredecessorOfEnd, InteractionState interactionState) {
     reorderInteractionFragment(scenario, newPredecessor, interactionState);
   }
-  
+
   private static void reorderInteractionOperand(Scenario scenario, InteractionFragment newPredecessor,
-    InteractionFragment newPredecessorOfEnd, InteractionOperand interactionOperand) {
+      InteractionFragment newPredecessorOfEnd, InteractionOperand interactionOperand) {
     reorderInteractionFragment(scenario, newPredecessor, interactionOperand);
   }
 
@@ -329,7 +258,7 @@ public class SequenceDiagramServices {
       InteractionFragment newPredecessorOfEnd, SequenceMessage reorderedMessage) {
     AbstractEnd begin = reorderedMessage.getSendingEnd();
     AbstractEnd end = reorderedMessage.getReceivingEnd();
-    
+
     if (begin != null) {
       reorderInteractionFragment(scenario, newPredecessor, begin);
     }
@@ -343,7 +272,7 @@ public class SequenceDiagramServices {
     fragments = scenario.getOwnedInteractionFragments();
     messages = scenario.getOwnedMessages();
 
-    //Reorder shall use move instead of remove/add
+    // Reorder shall use move instead of remove/add
     if (newPredecessor == null) {
       // The message is pushed up to the top of the message list
       messages.move(0, reorderedMessage);
@@ -352,10 +281,10 @@ public class SequenceDiagramServices {
       SequenceMessage previousMessage = findCorrespondingPreviousMessage(fragments, newPredecessor);
       if (previousMessage != null) {
         int positionMessage = messages.indexOf(previousMessage);
-        if (positionMessage +1 <messages.size()){
-            messages.move(positionMessage +1, reorderedMessage);
+        if (positionMessage + 1 < messages.size()) {
+          messages.move(positionMessage + 1, reorderedMessage);
         } else {
-            messages.move(positionMessage, reorderedMessage);
+          messages.move(positionMessage, reorderedMessage);
         }
       } else {
         // The message is pushed up to the top of the message list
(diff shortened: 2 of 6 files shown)


core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/refresh/extension/InteractionRefreshExtension.java
```diff
@@ -0,0 +1,158 @@
...
+   * Cache for InstanceRole -> EventContext structure.
+   */
+  private static Map<InstanceRole, List<EventContext>> instanceRoleToEventContextCache = new ConcurrentHashMap<InstanceRole, List<EventContext>>();
+
+  /**
+   * @see org.eclipse.sirius.business.api.refresh.IRefreshExtension#beforeRefresh(org.eclipse.sirius.DDiagram)
+   */
+  @Override
+  public void beforeRefresh(DDiagram diagram) {
+    super.beforeRefresh(diagram);
```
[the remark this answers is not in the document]
mPorhel: cacheGetter could return null is cache is not enabled and put method would do nothing (but current implementations which do put and then get would need to be reworked)

mPorhel (changes requested):

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/refresh/extension/InteractionRefreshExtensionProvider.java
```diff
@@ -19,24 +19,24 @@
...
-    /**
-     * @see org.eclipse.sirius.business.api.refresh.IRefreshExtensionProvider#provides(org.eclipse.sirius.DDiagram)
-     */
-    public boolean provides(DDiagram viewPoint_p) {
-        return viewPoint_p instanceof SequenceDDiagram;
-    }
+  /**
+   * @see org.eclipse.sirius.business.api.refresh.IRefreshExtensionProvider#provides(org.eclipse.sirius.DDiagram)
+   */
+  public boolean provides(DDiagram viewPoint) {
```
mPorhel, line 38: diagram ?

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/refresh/extension/InteractionRefreshExtension.java
```diff
@@ -65,7 +69,7 @@ public void beforeRefresh(DDiagram diagram) {
   @Override
   public void postRefresh(DDiagram diagram) {
     clearCaches();
-    super.postRefresh(diagram);
+    isCacheEnabled = false;
```
mPorhel, line 72: We could condition the putxxxCache method to have an effect only in this instead of check theisCacheEnabled value everywhere.

Teh need to have the isCacheEnabledMethod would be less usefull.

mPorhel (commented):
See also my previous remarks on putEndToEventCahce in visit() methods. 

It seems that a lot of other .putXxxXxxCache method still used without checking the cache status
We need to check from where the services are used : which tools ? (creation tools which use a CreateView operation for exemple will use the service outside the refresh : one creation tool for a synccall might lead to the creation of an execution + 2 edges) 

The getEndToEventCache should not been used to place/register the execution which sends/receive a message (l242) it should be an information from the EventContext structure.

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/refresh/extension/InteractionRefreshExtension.java
mPorhel: If we remove the super.beforeRefresh / afterRefresh method could you confirm that there is another extension which enable the ModelCache during the SequenceDDiagram refresh ? 

If there is another extension doing this, we can remove the superbeforeR/afterR methods but in this case we should also remove the inheritance to the AbstractCacheAwareRefreshExtension.

It seems that the TitleBlockRefreshExtensionProvider is enabled for every diagram.

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
```diff
@@ -0,0 +1,377 @@
...
+    // get executions from cache
+    Collection<EObject> directEvents = getElementToContainerCache(element);
+    if (directEvents != null) {
+      return directEvents.stream().filter(Execution.class::isInstance).map(Execution.class::cast)
+          .collect(Collectors.toList());
+    }
+    // if not present, compute it and put it in cache
+    return getDirectEvents(element, instanceRole).stream().filter(Execution.class::isInstance)
+        .map(Execution.class::cast).collect(Collectors.toList());
+  }
```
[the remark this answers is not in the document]
mPorhel: Done

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
```diff
@@ -0,0 +1,377 @@
...
+    // get states from cache
+    Collection<EObject> directEvents = getElementToContainerCache(element);
+    if (directEvents != null) {
+      return directEvents.stream().filter(StateFragment.class::isInstance).map(StateFragment.class::cast)
+          .collect(Collectors.toList());
+    }
+    // if not present, compute it and put it in cache
+    return getDirectEvents(element, instanceRole).stream().filter(StateFragment.class::isInstance)
+        .map(StateFragment.class::cast).collect(Collectors.toList());
+  }
```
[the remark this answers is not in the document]
mPorhel: Done

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
```diff
@@ -0,0 +1,377 @@
...
+   */
+  private static void computeTimeLapseStartAndEndCache(Scenario scenario) {
+    // cache Execution.start -> Execution and Execution.finish -> Execution
+    Stream<TimeLapse> timeLapses = scenario.getOwnedTimeLapses().stream()
+        .filter(tl -> !(tl instanceof AbstractFragment));
+    // scan timelapse only one time
+    if (getEndToEventCache().isEmpty()) {
+      timeLapses.forEach(e -> {
+        if (e.getStart() instanceof MessageEnd) {
+          putEndToEventCache(e.getStart(), e);
```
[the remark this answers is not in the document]
mPorhel: Done in further commit.

endToEventCache renamed into interactionFragmentToTimeLapseCache and type changed from Map<InteractionFragment, CapellaElement> to Map<InteractionFragment, TimeLapse>.

This cache is now used only to store and retrieve the link from an InteractionFragment to a TimeLapse, this information exists with ExecutionEnd::execution but is missing between MEssageEnd and Execution and between InteractionState and StateFragment. The other way is defined in the M2 with TimeLapse::start and TimeLapse::finish references.

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
```diff
@@ -0,0 +1,377 @@
...
+  /**
+   * Return abstract end covered TimeLapse/InstanceRole.
+   * 
+   * @param end
+   *          AbstractEnd
+   * @param instanceRole
+   *          InstanceRole
+   * @return TimeLapse
+   */
+  public static Optional<CapellaElement> getEvent(AbstractEnd end, InstanceRole instanceRole) {
```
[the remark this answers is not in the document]
mPorhel: Replace by computation of ReceivingEnd/SendingEnd based on the EventContext structure

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
```diff
@@ -0,0 +1,377 @@
...
+      ancestors.push(container.get());
+    }
+
+    if (container.isPresent() && container.get() instanceof Execution
+        && ((Execution) container.get()).getFinish() == end) {
+      ancestors.pop();
+      result.add(new EventContext(ancestors.peek(), container.get(), false, ancestors.size() + 1));
+    }
+    if (!container.isPresent()) {
+      putEndToEventCache(end, (CapellaElement) ancestors.peek());
```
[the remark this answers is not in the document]
mPorhel: Replaced with the creation of EventContext for every MessageEnd, see comments in code.

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
```diff
@@ -0,0 +1,377 @@
...
+        && ((StateFragment) container.get()).getStart() == end) {
+      result.add(new EventContext(ancestors.peek(), container.get(), true, ancestors.size() + 1));
+    }
+
+    if (container.isPresent() && container.get() instanceof StateFragment
+        && ((StateFragment) container.get()).getFinish() == end) {
+      result.add(new EventContext(ancestors.peek(), container.get(), false, ancestors.size() + 1));
+    }
+    if (!container.isPresent()) {
+      putEndToEventCache(end, (CapellaElement) ancestors.peek());
```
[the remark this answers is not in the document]
mPorhel: LoC removed.

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/SequenceDiagramServices.java
```diff
@@ -141,69 +113,21 @@ public static EObject getSendingEnd(SequenceMessage message) {
...
-    Scenario sc = (Scenario) message.eContainer();
-    for (TimeLapse exec : sc.getOwnedTimeLapses()) {
-      if (exec.getFinish() == end) {
-        return exec;
-      }
-    }
-    Execution execution = getExecutionOfMessageEnd(end);
-
-    return execution == null ? end.getCovered() : execution;
+    return getExecutionOrCoveredOfMessageEnd(end);
```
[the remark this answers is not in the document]
mPorhel: Computation is now based on EventContext structure.

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/SequenceDiagramServices.java
```diff
@@ -141,69 +113,21 @@ public static EObject getSendingEnd(SequenceMessage message) {
...
-            return top(executionStack);
-          }
-        }
-      }
-    }
-    return null;
+  private static EObject getExecutionOrCoveredOfMessageEnd(MessageEnd end) {
+    Optional<CapellaElement> execution = EventContextServices.getEvent(end, currentInstanceRole(end));
+    // null never happen
+    return execution.isPresent() ? execution.get() : null;
```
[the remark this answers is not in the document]
mPorhel: Computation is now based on EventContext structure.

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/refresh/extension/InteractionRefreshExtension.java
```diff
@@ -0,0 +1,158 @@
...
+   * Cache for MessageEnd -> Execution/InstanceRole and InteractionState -> StateFragment. (Cross referencer might be a
+   * way to get this information.)
+   */
+  private static Map<InteractionFragment, CapellaElement> endToEventCache = new ConcurrentHashMap<>();
+
+  /**
+   * Cache for ChildExecution -> ParentExecution and StateFragment -> ParentExecution.
+   */
+  private static Map<EObject, Collection<EObject>> elementToContainerCache = new ConcurrentHashMap<>();
+
```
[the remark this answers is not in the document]
mPorhel: elementToContainerCache renamed into directEventsCache.
It could have been named directChildrenTimeLapseCache.

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/refresh/extension/InteractionRefreshExtension.java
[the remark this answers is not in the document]
mPorhel: Done

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/refresh/extension/InteractionRefreshExtension.java
```diff
@@ -0,0 +1,158 @@
...
+  public void beforeRefresh(DDiagram diagram) {
+    super.beforeRefresh(diagram);
+  }
+
+  /**
+   * @see org.eclipse.sirius.business.api.refresh.IRefreshExtension#postRefresh(org.eclipse.sirius.DDiagram)
+   */
+  @Override
+  public void postRefresh(DDiagram diagram) {
+    interactionCache.clearCache();
```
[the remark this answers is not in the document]
mPorhel: done

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/refresh/extension/InteractionRefreshExtensionProvider.java
mPorhel: done

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/refresh/extension/InteractionRefreshExtension.java
mPorhel: done

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
```diff
@@ -0,0 +1,414 @@
...
+   * 
+   * @param scenario
+   *          Scenario
+   */
+  private static void computeTimeLapseStartAndEndCache(Scenario scenario) {
+    // cache Execution.start -> Execution and Execution.finish -> Execution
+    // scan timelapse only one time
+    if (getInteractionFragmentToTimeLapseCache().isEmpty()) {
+      Stream<TimeLapse> timeLapses = scenario.getOwnedTimeLapses().stream()
+          .filter(tl -> !(tl instanceof AbstractFragment));
```
minhtutonthat, line 358: Maybe AbstractFragment could be taken into account in a future enhancement since it's absolutely suitable to use the same logic of EventContext to process elements such as CombinedFragment and InteractionUse.

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
```diff
@@ -0,0 +1,414 @@
...
+        //        |         |         - 
+        //        |         |         | 
+        
+        // would result the following structures:
+        
+        //  For IR1: 
+        //   EventContext(IR1, m1, true, 1)
+        //   EventContext(IR1, return_m1, false, 1)
+        //
+        //  For IR1:
```
minhtutonthat, line 264: For IR2?

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/SequenceDiagramServices.java
```diff
@@ -140,70 +111,28 @@ public static EObject getSendingEnd(SequenceMessage message) {
     MessageEnd end = message.getSendingEnd();
     if (end == null) {
       return message; // found message case
-    }
+    } else {
```
minhtutonthat, line 114: Could you remove the unnecessary else clause?

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/SequenceDiagramServices.java
```diff
@@ -140,70 +111,28 @@ public static EObject getSendingEnd(SequenceMessage message) {
...
-    Scenario sc = (Scenario) message.eContainer();
-    for (TimeLapse exec : sc.getOwnedTimeLapses()) {
-      if (exec.getStart() == end) {
-        return exec;
-      }
-    }
-    Execution execution = getExecutionOfMessageEnd(end);
-    return execution == null ? end.getCovered() : execution;
-  }
+    } else {
```
minhtutonthat, line 128: Similary, could you remove the else clause plz?

minhtutonthat (dismissed):

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sequencediag/ScenarioService.java
```diff
@@ -88,6 +86,7 @@
 import org.polarsys.capella.core.model.helpers.queries.QueryIdentifierConstants;
 import org.polarsys.capella.core.model.helpers.queries.filters.OnlySharedDataOrEventOrUnsetFilter;
 import org.polarsys.capella.core.sirius.analysis.CapellaServices;
+import org.polarsys.capella.core.sirius.analysis.EventContextServices;
```
sandupostaru, line 89: **Preambule:**
Thank you for the contribution. The new caching system seems cool and I can see the benefits from this change :+1: 

My comments ones are mostly related to style, class location, variable naming conventions and are not all that functional except for a few. Don't hesitate to come back to me if you need more details.

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
```diff
@@ -0,0 +1,414 @@
...
+      });
+    }
+  }
+
+  /**
+   * Helper class to keep track of who "contains" who depending on the interleaving of the start/finish ends.
+   * 
+   * @author pcdavid
+   */
+  public static final class EventContext {
```
sandupostaru, line 376: The name of this class is confusing. There is already an `Event` class in Capella that has different meaning.

A more suitable name  would be `Sibling` because this class contains siblings of InstanceRoles (closely related elements).
An sibling contains a `parent` which can be used to filter the `children`.

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
```diff
@@ -0,0 +1,414 @@
...
+    private final EObject parent;
+
+    private final boolean start;
+
+    private final EObject element;
```
sandupostaru, line 381: change to `child`

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
```diff
@@ -0,0 +1,414 @@
...
+
+    public boolean isStart() {
+      return start;
+    }
+
+    public EObject getParent() {
+      return parent;
+    }
+
+    public EObject getElement() {
```
sandupostaru, line 400: getChild

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
```diff
@@ -0,0 +1,414 @@
...
+  }
+
+  /**
+   * Compute EventContext structure from InstanceRole if needed and put it in cache.
+   * 
+   * @param instanceRole
+   *          InstanceRole
+   * @return List<EventContext>
+   */
+  public static List<EventContext> getEventContexts(InstanceRole instanceRole) {
```
sandupostaru, line 69: rename to `getSiblings` (see my comment on the renaming of the `EventContext` class)

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
```diff
@@ -0,0 +1,414 @@
...
+   * Get direct event on element. (InstanceRole -> Executions + StateFragment, Execution -> Execution children +
+   * StateFragment)
+   * 
+   * @param element
+   *          EObject
+   * @param instanceRole
+   *          InstanceRole
+   * @return Collection<EObject>
+   */
+  public static Collection<EObject> getDirectEvents(EObject element, InstanceRole instanceRole) {
```
sandupostaru, line 90: make this method `private` and rename to `getSiblingChildren`

sandupostaru, line 90: Rename `element` to `parent` and change the order of the parameters, `instanceRole` comes first, `parent` second

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
```diff
@@ -0,0 +1,414 @@
...
+   *          InstanceRole
+   * @return Collection<EObject>
+   */
+  public static Collection<Execution> getExecutionDirectEvents(EObject element, InstanceRole instanceRole) {
+    // get executions from cache
+    Collection<EObject> directEvents = getDirectEventsFromCache(element);
+    if (directEvents != null) {
+      return directEvents.stream().filter(Execution.class::isInstance).map(Execution.class::cast)
+          .collect(Collectors.toList());
+    }
```
sandupostaru, line 120: This code block  does not serve any purpose, since the `getDirectEvents()` at line 122 already does verify if the information is available in the cache. This block should thus be removed.

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
```diff
@@ -0,0 +1,414 @@
...
+    // get states from cache
+    Collection<EObject> directEvents = getDirectEventsFromCache(element);
+    if (directEvents != null) {
+      return directEvents.stream().filter(StateFragment.class::isInstance).map(StateFragment.class::cast)
+          .collect(Collectors.toList());
+    }
+    // if not present, compute it and put it in cache
+    return getDirectEvents(element, instanceRole).stream().filter(StateFragment.class::isInstance)
+        .map(StateFragment.class::cast).collect(Collectors.toList());
+  }
```
sandupostaru, line 145: This code block  does not serve any purpose, since the `getDirectEvents()` at line 143 already does verify if the information is available in the cache. This block should thus be removed.

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
```diff
@@ -0,0 +1,414 @@
...
+  /**
+   * Get Execution direct events on element.
+   * 
+   * @param element
+   *          EObject
+   * @param instanceRole
+   *          InstanceRole
+   * @return Collection<EObject>
+   */
+  public static Collection<Execution> getExecutionDirectEvents(EObject element, InstanceRole instanceRole) {
```
sandupostaru, line 114: rename to `getSiblingExecutions`, change parameter order and rename `element` to `parent`

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
```diff
@@ -0,0 +1,414 @@
...
+  /**
+   * Get StateFragment direct events on element.
+   * 
+   * @param element
+   *          EObject
+   * @param instanceRole
+   *          InstanceRole
+   * @return Collection<EObject>
+   */
+  public static List<StateFragment> getStateFragmentDirectEvents(EObject element, InstanceRole instanceRole) {
```
sandupostaru, line 135: rename to `getSiblingStateFragments`, change parameter order and rename `element` to `parent`

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
```diff
@@ -0,0 +1,414 @@
...
+  }
+
+  /**
+   * Compute EventContext structure for InstanceRole
+   * 
+   * @param instanceRole
+   *          InstanceRole
+   * @return List<EventContext>
+   */
+  public static List<EventContext> computeInstanceRoleEventContextStructure(InstanceRole instanceRole) {
```
sandupostaru, line 154: rename to `computeSiblings`

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
```diff
@@ -0,0 +1,414 @@
...
+    if (instanceRole == null || !(instanceRole.eContainer() instanceof Scenario)) {
+      return Collections.emptyList();
+    }
+    Scenario scenario = SequenceDiagramServices.getScenario(instanceRole);
+
+    // initialize ancestors
+    Deque<CapellaElement> ancestors = new ArrayDeque<>();
+    ancestors.push(instanceRole);
+    List<EventContext> result = new ArrayList<>();
```
sandupostaru, line 163: rename to `siblings`

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
```diff
@@ -0,0 +1,414 @@
...
+
+    // Cache missing info from M2 :
+    // - cache Execution.start -> Execution and Execution.finish -> Execution when the end is a MessageEnd
+    // - cache StateFragment.start -> InteractionState and StateFragment.finish -> InteractionState
+    // (ExecutionEnd has a getExecution() method)(ExecutionEnd has a getExecution() method)
+    computeTimeLapseStartAndEndCache(scenario);
+
+    // compute Execution/StateFragment and InteractionFragment structure
+    Stream<InteractionFragment> ends = scenario.getOwnedInteractionFragments().stream()
```
sandupostaru, line 172: rename to `interactionFragments`

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
```diff
@@ -0,0 +1,414 @@
...
+        .filter(frag -> frag instanceof AbstractEnd || frag instanceof InteractionState);
+    ends.forEachOrdered(end -> {
```
sandupostaru, line 174: rename to `interactionFragment`

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
```diff
@@ -0,0 +1,414 @@
...
+      InstanceRole covered = getCoveredInstanceRole(end);
```
sandupostaru, line 175: `coveredInstanceRole`

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
```diff
@@ -0,0 +1,414 @@
...
+      if (covered != null && covered.equals(instanceRole)) {
+        Optional<TimeLapse> timeLapse = getTimeLapseFromCache(end);
```
sandupostaru, line 177: check right away if `timelapse` is null / present here and thus avoid checking it in each of the visitMethods

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
```diff
@@ -0,0 +1,414 @@
...
+
+        // Interaction State case
+        if (end instanceof InteractionState) {
+          visit(ancestors, result, (InteractionState) end, timeLapse);
+        }
+
+        // Message End case
+        if (end instanceof MessageEnd) {
+          visit(ancestors, result, (MessageEnd) end, timeLapse);
+        }
```
sandupostaru, line 192: See the comments for each of the `visit` methods

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
```diff
@@ -0,0 +1,414 @@
...
+   * Compute EventStructure for MessageEnd.
+   * 
+   * @param ancestors
+   *          Deque<EObject>
+   * @param result
+   *          List<EventContext>
+   * @param end
+   *          MessageEnd
+   */
+  private static void visit(Deque<CapellaElement> ancestors, List<EventContext> result, MessageEnd end,
```
sandupostaru, line 209: rename to ComputeSiblings

Change the order of the arguments and their name like so:

`computeSiblings(ancestors, timelapse->child , end -> executionEnd/state/messageEnd [depending on the method], result -> siblings)`

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
```diff
@@ -0,0 +1,414 @@
...
+   * Compute EventStructure for InteractionState.
+   * 
+   * @param ancestors
+   *          Deque<EObject>
+   * @param result
+   *          List<EventContext>
+   * @param end
+   *          InteractionState
+   */
+  private static void visit(Deque<CapellaElement> ancestors, List<EventContext> result, InteractionState end,
```
sandupostaru, line 292: same remark as above

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
```diff
@@ -0,0 +1,414 @@
...
+   * Compute EventStructure for ExecutionEnd.
+   * 
+   * @param ancestors
+   *          Deque<EObject>
+   * @param result
+   *          List<EventContext>
+   * @param end
+   *          ExecutionEnd
+   */
+  private static void visit(Deque<CapellaElement> ancestors, List<EventContext> result, ExecutionEnd end,
```
sandupostaru, line 315: same remark as above

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
```diff
@@ -0,0 +1,414 @@
...
+   * @param result
+   *          List<EventContext>
+   * @param end
+   *          MessageEnd
+   */
+  private static void visit(Deque<CapellaElement> ancestors, List<EventContext> result, MessageEnd end,
+      Optional<TimeLapse> timeLapse) {
+
+    // Execution start / SyncCall and ASyncCall
+    if (timeLapse.isPresent() && timeLapse.get() instanceof Execution
```
sandupostaru, line 213: not needed since null check is moved in calling method

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
```diff
@@ -0,0 +1,414 @@
...
+
+    // Execution with return branch end / SyncCall
+    if (timeLapse.isPresent() && timeLapse.get() instanceof Execution
+        && ((Execution) timeLapse.get()).getFinish() == end) {
+      ancestors.pop();
+      result.add(new EventContext(ancestors.peek(), timeLapse.get(), false, ancestors.size() + 1));
+    }
+
+    // @formatter:off
+        //  The Following diagram would result in
```
sandupostaru, line 241: Thanks for the explication, this is nice :+1:

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/refresh/extension/InteractionRefreshExtension.java
```diff
@@ -0,0 +1,188 @@
...
+  public static void putDirectEventsInCache(EObject element, Collection<EObject> container) {
+    directEventsCache.put(element, container);
+  }
+
+  /**
+   * @param key
+   *          InteractionFragment
+   * @return Event corresponding to key.
+   */
+  public static Optional<TimeLapse> getTimeLapseFromCache(InteractionFragment key) {
```
sandupostaru, line 145: Why does this method return an optional? All of the other methods return null if the value is absent. 
Let's return a `TimeLapse` here ;)

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/refresh/extension/InteractionRefreshExtensionProvider.java
```diff
@@ -0,0 +1,42 @@
...
+ *    Thales - initial API and implementation
+ *******************************************************************************/
+package org.polarsys.capella.core.sirius.analysis.refresh.extension;
+
+import org.eclipse.sirius.diagram.DDiagram;
+import org.eclipse.sirius.diagram.business.api.refresh.IRefreshExtension;
+import org.eclipse.sirius.diagram.business.api.refresh.IRefreshExtensionProvider;
+import org.eclipse.sirius.diagram.sequence.SequenceDDiagram;
+
+public class InteractionRefreshExtensionProvider implements IRefreshExtensionProvider {
```
sandupostaru, line 20: Lets call this Class `ScenarioRefreshExtensionProvider` in order to be coherent.

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/refresh/extension/InteractionRefreshExtension.java
```diff
@@ -0,0 +1,188 @@
...
+import org.eclipse.emf.ecore.EObject;
+import org.eclipse.sirius.diagram.DDiagram;
+import org.eclipse.sirius.diagram.business.api.refresh.IRefreshExtension;
+import org.polarsys.capella.core.data.helpers.cache.Cache;
+import org.polarsys.capella.core.data.interaction.InstanceRole;
+import org.polarsys.capella.core.data.interaction.InteractionFragment;
+import org.polarsys.capella.core.data.interaction.TimeLapse;
+import org.polarsys.capella.core.sirius.analysis.EventContextServices.EventContext;
+
+public class InteractionRefreshExtension implements IRefreshExtension {
```
sandupostaru, line 31: Lets call this Class `ScenarioRefreshExtension` in order to be coherent.

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/refresh/extension/InteractionRefreshExtension.java
```diff
@@ -0,0 +1,188 @@
...
+   * Is InteractionRefreshExtension caches enabled ?
+   */
+  private static boolean isCacheEnabled;
+
+  /**
+   * @see org.eclipse.sirius.business.api.refresh.IRefreshExtension#beforeRefresh(org.eclipse.sirius.DDiagram)
+   */
+  @Override
+  public void beforeRefresh(DDiagram diagram) {
+      enableRefreshCache();
```
sandupostaru, line 64: We should also call `clearCache()` here, in case the root `postRefresh` Sirius method fails for whatever reason, and our `postRefresh` is not called.

sandupostaru (changes requested):

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
mPorhel: Hi Sandu, 
We kept Event,  EventContext and EventEnd as this is has a meaning in the Sirius Sequence diagram (used in layout, internal api ISequenceEvent), it was really helpfull in debug to have the same wording than in Sirius (refresh, layout) and not new vocabulary. 
It was also a way to be sure to avoid to reuse TimeLapse, Fragments, ...

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
mPorhel: I am not sure, this is really the current element for which we want the parent. 
Child would be used if we wanted to compute the children of element.

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
mPorhel: It was hard ten years after the main developments on Sequence to bring all memories together, it will be easier with this.

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/refresh/extension/InteractionRefreshExtension.java
mPorhel: Because it is a special case :

 it is used  to simulate the "MixEnd" case : some MessageEnd are used as start/finish of an Execution additionnaly to being the start/finish of a message. But there is no end -> Execution/TimeLapse info in the M2.  
We also use this cache to retrieve the InteractionState > StateFragment information.
And to be consistent, we also store ExecutionEnd -> Execution. 

If the cache returns null, IMO, this means that we put no entry in it. If the cache returns en empty Optional, this means that the TimeLapse in the model has no start or finish. 

But looking at the implementation, it seems that putTimeLapseInCache should do nothing if the provided key is null and you are right getTimeLapseFromCache could return a TimeLapse, without Optional.

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
mPorhel: this code filters the directEvents and return only the executions.

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/SequenceDiagramServices.java
mPorhel: In our Dev environment, without the "else", we have a Checsktyle warning.

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
mPorhel: Hi Sandu. 

After discussion with Philippe, it is ok to rename the EventContext into something else, to avoid confusion with the "event" from the Capella M2 to help the maintainers of this code, but "sibling" does not look appropriate: 

> definition of sibling: each of two or more children or offspring having one or both parents in common; a brother or sister.

IMO, we should try to find another name, as the EventContext elements are not just siblings under the same parent, but their organization represents the hierarchical tree of Executions/States on the diagram and the structure is used from the VSM to build/find the semantic candidates for level/depth in evaluation of the mappings's expression during the SequenceDDiagram refresh.

mPorhel:
Hi @sandupostaru , 

Nathalie will provide a new commit with EventContextServices.EventContex renamed into SequenceCache.SemanticCandidateContext.

Regards

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/cache/ScenarioCache.java
```diff
@@ -0,0 +1,554 @@
...
+    // | | | | |
+    // | | - <-----| | e3 m3
+    // | || | | |
+    // | || | | |
+    // | | - ----->| | return_m3
+    // | | | | |
+    // |<------- - | | return_m1
+    // | | -
+    // | | |
+
```
sandupostaru, line 399: The formatting of this section seems to be off. I think it was messed-up during the save + auto format.

Can you restore it please?

sandupostaru (changes requested):

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/EventContextServices.java
minhtutonthat: Ok i see

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/SequenceDiagramServices.java
minhtutonthat: In our dev environment Sonar checks this kind of rule :)

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/cache/ScenarioCache.java
mPorhel: Hi Sandu,
Nathalie is on holidays until Monday. 
She will fix this on Monday. 
We need to disable the Eclipse formatter  for this section as it is the only solution to have a understandable view of the computation here.

nlepine: Done!

sandupostaru (approved):

sandupostaru:
LGTM

PR #384, opened 2020-10-02
 [567469] Transition of referenced Scenarios

georgiana-ecobici:


core/plugins/org.polarsys.capella.core.data.interaction.ui.quickfix/plugin.properties
```diff
@@ -1,5 +1,5 @@
 # =============================================================================
-# Copyright (c) 2006, 2015 THALES GLOBAL SERVICES.
+# Copyright (c) 2006, 2020 THALES GLOBAL SERVICES.
```
pdulth, line 2: you don't have to bother you to keep copyright date up-to-date, for thales ones, we do it once in a year (for instance, https://github.com/eclipse/capella/commit/22271eb84f0539e58781bdbb11fa2544e271e307)

PR #404, opened 2020-10-15
[567541] Make "refresh all sub representation" more reliable

lfasani:
- close all editors before refreshing all representation otherwise the
opened editors won't be refreshed.
- report properly when the refresh of a representation failed. That may
happen if an exception is raised either in OpeningDiagramJob or
indirectly in the refresh done on the UI thread.
- add log info in "Information" view
- log only a summary in error log view

Change-Id: I348c80164a2fcea38d644459de523faab10fd4fe


pdulth:
Merged by #419

PR #449, opened 2020-11-25
[569167] Capability transition does not work for involved components/actors and other capability relations

minhtutonthat:
Bug: 569167
Change-Id: I2c8ca3a6a67a39270cffd19ac2b32bb48c640351


pdulth (approved):
Looks good, not a simple fix..

Issue #2024, opened 2021-01-11
CUSTOMER - Constraints on OEntities and OActors should be displayed in referencing elements inside the Semantic Browser view 


claudiasima-ro:
Constraints applied on the selected element should be displayed in the Semantic Browser (left-hand part: Constraining Elements).
They are not displayed on Operational Entities. Same problem for Operational Actors (Operational Entity [Actor]).

Note that they are correctly displayed on System Actors, the System, Operational Activities and Interactions.

PR #2025, opened 2021-01-11
#2024 Constraints on OEntities and OActors should be displayed in

claudiasima-ro:
referencing elements inside the Semantic Browser view

### DIFF
--- a/core/plugins/org.polarsys.capella.core.semantic.queries/src/org/polarsys/capella/core/semantic/queries/basic/queries/ModelElementConstraints.java
+++ b/core/plugins/org.polarsys.capella.core.semantic.queries/src/org/polarsys/capella/core/semantic/queries/basic/queries/ModelElementConstraints.java
@@ -77,6 +77,9 @@ public class ModelElementConstraints implements IQuery {
       for (AbstractConstraint constraint : part.getConstraints())
         result.add(constraint);
     }
+    for (AbstractConstraint constraint : component.getConstraints()) {
+      result.add(constraint);
+    }
 
     return result;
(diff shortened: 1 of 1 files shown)


claudiasima-ro:
This issue was fixed for OEntities.
For OActors, and also for Physical Components of type Deploy (Node or Behavior), the issue will be fixed in another story, as it requires more analysis.

core/plugins/org.polarsys.capella.core.semantic.queries/src/org/polarsys/capella/core/semantic/queries/basic/queries/ModelElementConstraints.java
```diff
@@ -54,6 +55,8 @@ public ModelElementConstraints() {
     List<Object> result = new ArrayList<Object>();
     if (object instanceof ModelElement) {
       EList<AbstractConstraint> constraints = ECollections.newBasicEList();
+      if (object instanceof Entity)
```
minhtutonthat, line 58: This is not the good fix. The original problem is rather on the creation of constraint. When you create a constraint and associate it to an Entity or an OA, it associates to the Entity/OA but not the part. That's why this query does not return anything. When you do the same thing but for a Logical Component for example, you'll see that the constraint is properly associated to the part and when you click on the Logical Component, you'll see the constraining elements.

minhtutonthat (changes requested):

georgiana-ecobici (approved):

minhtutonthat (approved):

Issue #2038, opened 2021-01-18
 New validation rule to detect Mix of Mode and State in State Machine


MalinaStefaniaStoicanescu:
Add new validation rule to detect Mix of Mode and State in Hierarchy and update QF for the actual one.

PR #2039, opened 2021-01-18
#2038 New validation rule to detect Mix of Mode/State in StateMachine

MalinaStefaniaStoicanescu:
- [x] Update QF for old rule + change 'warning' to 'error'
- [x] Add a new rule to detect mixed modes/states in the same state machine (different region)

Change-Id: Iadf8e9da0ea0398f5c6532c68a2ca6d07713e0ae
Signed-off-by: MalinaStefaniaStoicanescu <malina.stoicanescu@thalesgroup.com>


core/plugins/org.polarsys.capella.core.data.common.validation/src/org/polarsys/capella/core/data/common/validation/statetransition/MDCHK_State_Mode_Siblings.java
```diff
@@ -0,0 +1,57 @@
+package org.polarsys.capella.core.data.common.validation.statetransition;
```
georgiana-ecobici, line 1: Add copyright header.

georgiana-ecobici (approved):

minhtutonthat (approved):

Issue #2034, opened 2021-01-18
D&D disable Mix of Mode and State in Hierarchy


MalinaStefaniaStoicanescu:
Drag and drop and Copy-Paste of region should not be possible in statemachine with modes in it. EXCEPT IF the related existing preference inside Capella/Model is used, i.e. mix M & S allowed

PR #2067, opened 2021-02-02
#2034 [cleanup] D&D from Region to Region from review

pdulth:
Change-Id: I4f7a85f73df5908e2cae8a7760d0b40ee17ec3d4


pdulth:
In the review of https://github.com/eclipse/capella/pull/2044, a review comment has been mistakenly forgotten i think.
https://github.com/eclipse/capella/pull/2044#discussion_r562460818
@MalinaStefaniaStoicanescu @minhtutonthat wdyt ? For me its also useless.

minhtutonthat:
Well it wasn't really mistakenly forgotten :P. It was because even though a region cannot be contained inside another region, the DnD still allows it and put it under the StateMachine. But you're right, that's related to another "bigger" bug when impossible DnDs are not forbidden. This check on Region should removed and the bug of forbidden DnD of should be fixed.

minhtutonthat (approved):

PR #2079, opened 2021-02-11
#2080 Display black internal links instead of many ones

pdulth:
Change-Id: I2eedb3a0e447d2fc9cc837c8c702a7f2e438489c


core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/PhysicalServices.java
```diff
@@ -1124,10 +1155,18 @@ public boolean isValidInternalLinkEdge(EdgeTarget currentSourceNode, EdgeTarget
...
+      return false;
     }
-    return true;
+    
+    DDiagram diagram = ((DDiagramElement) currentSourceNode).getParentDiagram();
+    DEdge anotherEdge = DiagramServices.getDiagramServices().findInternalLinkEdge(diagram, currentSourceNode,
+        currentTargetNode, getPhysicalPathInternLinkEdgeMapping(diagram));
+    boolean hasAnotherEdge = anotherEdge != null && !(path.equals(anotherEdge.getTarget()));
+    if (hasAnotherEdge) {
+      System.out.println();
```
minhtutonthat, line 1167: Could you remove this?

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/DiagramServices.java
```diff
@@ -1479,6 +1484,39 @@ public void selectElementInDiagram(DSemanticDecorator newTarget) {
    * @param mapping
    * @return
    */
+  public DEdge findInternalLinkEdge(DDiagram pDiagram, EdgeTarget sourceNode, EdgeTarget targetNode, EdgeMapping mapping) {
```
minhtutonthat, line 1487: Since source/target does not make a lot of sense here, could we just name the variables "firstPort" and "secondPort" (just an idea)? Could you also add a comment for this method that it is supposed to always return the same internal link between 2 ports regardless of the creation order of interal links ?

minhtutonthat (changes requested):
Could you also do a rebase? Thanks.

minhtutonthat (approved):

Issue #2122, opened 2021-03-11
Add org.apache.felix.gogo.shell to the target platform


GlennPlou:
An issue reported in T4C jobs and scheduler console was related to this bug: https://www.eclipse.org/forums/index.php/t/1105369/

The last version of org.apache.felix.gogo.shell 1.1.4.v20210111-1007 solves this problem, which is embedded in the last Orbit release https://download.eclipse.org/tools/orbit/downloads/drops/R20210223232630/repository

GlennPlou:
This issue cannot be resolved so the changes have been cancelled. The new version of org.apache.felix.gogo.shell has conflicting dependencies with the T4C configuration. The problem we faced is similar to https://www.eclipse.org/forums/index.php/t/1106342/

PR #2123, opened 2021-03-11
#2122 Add org.apache.felix.gogo.shell to the target platform

GlennPlou:


mPorhel:
Hi Sandu, Glenn, 

For the moment we consider that this PR should be canceled, GogoShell is brought by org.eclipse.e4.rcp which includes 1.1.0. 
That's why the result of this PR does not contain Gogo shell 1.1.4 (even if we complete the tp to get gogo runtime and other depenencies, it will not change the result).

GlennPlou:
NO GO for this PR: The ticket cannot be resolved due to a conflict of dependencies brought by org.apache.felix.gogo.shell.

PR #2157, opened 2021-04-01
#2151 Disable menus on empty selection

### LINKED ISSUE #2151: Visible menus on empty selectios
-  fix when validate menu and copy as menu is visible
-  fix copy as text on DRepresentarionDescriptor returned text
-  disable Migration and Open menus on empty selections
-  disable Validation menu if the selection from the action is disabled

georgiana-ecobici:
Resolves #2151

- Migrate, Open, Open and Show in Diagram Editor
- Disable Validate action when it contains empty selection

Change-Id: I6c7d64dcb5702a173a5bb5f5ef89df7e12e7e237
Signed-off-by: Georgiana Ecobici <georgiana-elena.ecobici@thalesgroup.com>

### DIFF
--- a/core/plugins/org.polarsys.capella.core.data.migration/plugin.xml
+++ b/core/plugins/org.polarsys.capella.core.data.migration/plugin.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <?eclipse version="3.2"?>
 <!--
-  Copyright (c) 2006, 2020 THALES GLOBAL SERVICES.
+  Copyright (c) 2006, 2021 THALES GLOBAL SERVICES.
   
   This program and the accompanying materials are made available under the
   terms of the Eclipse Public License 2.0 which is available at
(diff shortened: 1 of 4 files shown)


core/plugins/org.polarsys.capella.core.validation.ui/src/org/polarsys/capella/core/validation/ui/DynamicActionProvider.java
```diff
@@ -105,6 +105,8 @@ public ValidateAction createDefaultValidation() {
     ISelection selection = selectionProvider.getSelection();
     if (!selection.isEmpty()) {
       validationAction.selectionChanged(new SelectionChangedEvent(selectionProvider, selection));
+    } else {
+      validationAction.setEnabled(false);
     }
```
georgiana-ecobici, line 110: This change is for the case when an element was previously selected, but you click on empty space in SB (selection inside the validate action is empty).

image

core/plugins/org.polarsys.capella.core.data.migration/plugin.xml
```diff
@@ -90,6 +90,9 @@
                   commandId="org.polarsys.capella.core.data.migration.project">
                <visibleWhen
                      checkEnabled="true">
+		               <iterate
```
pdulth, line 93: image

pdulth (changes requested):
We will merge it for IT2,
The first fix makes the migration project displayed but disabled when we click on any other elements.

pdulth:
When you use one reference for the 3 commands, when clicking on an element, the non enabled will remain visible.

image

To have the same behavior than previous version, can you try to add 3 dedicated references, one per handler rather than one with the three handlers?
image
image

core/plugins/org.polarsys.capella.core.data.migration/src/org/polarsys/capella/core/data/migration/handlers/MigrationPropertyTester.java
```diff
@@ -0,0 +1,38 @@
...
+ * 
+ */
+public class MigrationPropertyTester extends PropertyTester {
+
+  @Override
+  public boolean test(Object receiver, String property, Object[] args, Object expectedValue) {
+    // check if is project, aird or model migration
+    switch(property) {
+    case "isValidMigrationProject":
+      return new ProjectMigrationHandler().isValidSelection(Arrays.asList(receiver));
```
pdulth: When we disable commands through "Window / Preferences / Capella / CommandConfigurability / Migration"
the handler on the command is replaced by another one by "PreferenceHandler.getInstance(command).activateHandler

If we use the "classical one" here, the command will be disabled and still visible even if we used a visibleWhen in the menu command.
In the new patchset, we retrieve the current command handler (classical one or PreferenceHelper disabling one) then look if it is enabled.

pdulth (approved):

Issue #2241, opened 2021-09-24
Ignore non-LibrariesPackage related UnresolvedReferenceException in LibrarySAXXMIHandler.handleForwardReferences


SteveMonnier:
Importing a shared Capella project where the root element (Project) has a status leads to an UnresolvedReferenceException that can be ignored in LibrarySAXXMIHandler.handleForwardReferences

SteveMonnier:
While I am trying to remember how to create a pull request, this fix only needs to add the method:


  @Override
  protected void handleForwardReferences(boolean isEndDocument) {
      super.handleForwardReferences(isEndDocument);
      List<Diagnostic> diagnosticToIgnore = new ArrayList<>();
      for (Diagnostic diagnostic : xmlResource.getErrors()) {
          if (diagnostic instanceof UnresolvedReferenceException && !(LibrariesPackage.eNS_URI.equals(((UnresolvedReferenceException) diagnostic).getFeature().eResource().getURI().toString()))) {
              diagnosticToIgnore.add(diagnostic);
          }
      }
      xmlResource.getErrors().removeAll(diagnosticToIgnore);
  }

PR #2243, opened 2021-09-28
#2241 Ignore non-LibrariesPackage related UnresolvedReferenceException

SteveMonnier:
Importing a shared Capella project where the root element (Project) has
a status leads to an UnresolvedReferenceException that can be ignored in
LibrarySAXXMIHandler.handleForwardReferences

Bug:2241
Change-Id: I5295d0fed0b247acf38aa2d5fcc3e6c29cb70d3d
Signed-off-by: Steve Monnier <steve.monnier@obeo.fr>


pdulth:
Closed, already merged https://github.com/eclipse/capella/blob/master-6.0/core/plugins/org.polarsys.capella.core.libraries/src/org/polarsys/capella/core/libraries/provider/LibrarySAXXMIHandler.java

PR #2284, opened 2021-11-29
[2278] Enable Migration from 1.4.x to 5.2

bzendagui:
- update the prevent migration process
- enhance the error message when the model version is too old, the model version is shown in the message

Change-Id: I11a35c68c4203c4e834cd0c78221afabdbf90dae
Signed-off-by: Boubekeur Zendagui <boubekeur.zendagui@thalesgroup.com>

### DIFF
--- a/core/plugins/org.polarsys.capella.core.data.migration/src/org/polarsys/capella/core/data/migration/af/ViewpointMigrationContribution.java
+++ b/core/plugins/org.polarsys.capella.core.data.migration/src/org/polarsys/capella/core/data/migration/af/ViewpointMigrationContribution.java
@@ -153,7 +153,7 @@ public class ViewpointMigrationContribution extends AbstractMigrationContributio
 
     if (!isMigrationPossible(fileVersion, currentVersion)) {
       String formattedMessage = NLS.bind(Messages.MigrationAction_ErrorDialog_TooOldMessage,
-          new String[] { context.getResourceName(), currentVersion.toString() });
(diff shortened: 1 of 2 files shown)


core/plugins/org.polarsys.capella.core.data.migration/src/org/polarsys/capella/core/data/migration/af/ViewpointMigrationContribution.java
```diff
@@ -194,17 +194,17 @@ public void postMigrationExecute(ExecutionManager executionManager, ResourceSet
...
-      }
-    }
-
-    return false;
+	// This allow migration from any 1.4.x version to 5.0, 5.1 and 5.2 to support the change of the versioning strategy (Issue #2278)
+	// This behavior should be removed from the migration process starting from the version 6.0
+	if ((fileVersion.getMajor() == 1 && fileVersion.getMinor() == 4) && (currentVersion.getMajor() == 5 && currentVersion.getMinor() <= 2)) {
+		return true;
+	}
+	return currentVersion.getMajor() - fileVersion.getMajor() <= 1;
```
pdulth, line 208: I this will allow migrations from too olds models like 1.3.x to 5.x which is not working. 

wdyt?
```
if (fileVersion.getMajor() == 1) 
    return fileVersion.getMinor() == 4) && (currentVersion.getMajor() == 5)
else 
    return currentVersion.getMajor() - fileVersion.getMajor() <= 1;
```

pdulth (changes requested):

pdulth (approved):

Issue #2273, opened 2021-11-17
Allow rotative image to use SVG files


arnauddieumegard:
Rotative workspace images cannot rely on SVG images this should be possible.

PR #2295, opened 2022-01-07
#2273 InFlowPort_left.svg is wrongly inverted

pdulth:


arnauddieumegard:
I may be mistaking but is it not InFlowPort_right.svg that is inverted ?

georgiana-ecobici (approved):

Issue #2303, opened 2022-01-11
Add computed transitions in M&S diagrams


nlepine:
As computed links functionality in Architecture Blank diagrams, add computed transitions in M&S diagrams. It displays transitions when sub regions, sub states or sub modes are hidden.
![image](https://user-images.githubusercontent.com/3184827/148934404-1e940c4a-3f47-4e9c-9a48-78bda6836102.png)

lredor:
Steps to validate:
* Import project MSMComputedTransitions from [MSMComputedTransitions.zip](https://github.com/eclipse/capella/files/8440861/MSMComputedTransitions.zip)
* Open diagram "[MSM] Default Region"
* Select "Mode 4" and "Mode 6"
* Hide them (Ctrl+H or menu "Show/Hide>Hide element")
* Expected: They are hidden. The transitions from or to them are also hidden.
* Launch a manual refresh
* Expected: "Computed transitions" are visible. They appear from or to the parent of the hidden modes.
* Select "Mode 2" and "Mode 8"
* Hide them (Ctrl+H or menu "Show/Hide>Hide element")
* Expected: They are hidden. The transitions from or to them are also hidden.
* Launch a manual refresh
* Expected: "Computed transitions" are visible. They appear from or to the parent of the hidden modes.
* Activate the filter "Hide Computed Transitions"
* Expected: The computed transitions are hidden.
* Disable the filter "Hide Computed Transitions"
* Expected: The computed transitions are visible.
* Activate the "Visibility Mode"
* Expected: The hidden elements are visible with "transparency".
* Select "Mode 2" and "Mode 4"
* Show them (menu "Show/Hide>Show element")
* Expected: They are visible. The transitions from or to them are also visible. The computed transitions are always visible.
* Launch a manual refresh
* Expected: Corresponding "computed transitions" are no longer visible.

PR #2305, opened 2022-01-11
#2303 Add computed transitions in M&S diagrams

nlepine:
Change-Id: I0bbc5dcd3d02272d1fb6196d9e98bc86c781b6d1


core/plugins/org.polarsys.capella.core.sirius.analysis/description/common.odesign
```diff
@@ -1389,7 +1399,7 @@
             </initialOperation>
           </ownedTools>
           <ownedTools xsi:type="tool:NodeCreationDescription" documentation="Creates a Entry Pseudo State" name="entry.point" label="%common.msm.entry.point" precondition="aql:self.canCreateEntryExitPoint(containerView)" forceRefresh="true" nodeMappings="//@ownedViewpoints[name='Common']/@ownedRepresentations[name='Mode%20State%20Machine']/@defaultLayer/@nodeMappings[name='MSM_ModeState']" iconPath="/org.polarsys.capella.core.sirius.analysis/icons/full/obj16/Pseudostate_entryPoint.gif">
-            <filters precondition="aql:true"/>
+            <filters precondition="aql:false"/>
```
pdulth, line 1402: I think the filter hides the tools because the tools are not working properly.
The entry point shall be created under a state and not the first region. 
Can you create a new PR dedicated to the entry/exit points changes as the mapping and tools need to be changed too

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/ModeStateMachineServices.java
```diff
@@ -170,6 +175,61 @@ public EObject showHideStatesInStateAndModeDiag(DSemanticDecorator view, List<St
...
+    while (contentIte.hasNext()) {
+      EObject next = contentIte.next();
+
+      if (next instanceof StateTransition) {
+        StateTransition transition = (StateTransition) next;
+
+        AbstractState source = transition.getSource();
+
+        AbstractDNode sourceRep = getMappingMSMStateMode(crossRef, source, diagram);
+        if (sourceRep != null && !sourceRep.isVisible()) {
```
pdulth, line 192: here, i think the filter doesn't work when the user use an "unsynchronized" diagram.
In an unsyncrhoznied diagram, all the sub modes are not displayed automatically but can be inserted/removed by the user.

(in capella tests, and more generally in capella code, the "show/hide keywords" reflects this behavior (more an insert/remove) and not the show/hide sirius feature which is using the setVisible (i.e. the MSMdiagram.hideState will "remove" the mode from the diagram, not "sirius hide" it.)

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/ModeStateMachineServices.java
```diff
@@ -336,6 +396,30 @@ public EObject showHideMSMTransitions(DSemanticDecorator view, List<StateTransit
     return view;
   }
 
+  private boolean isIn(AbstractDNode node, DDiagram diag) {
+    return node.getParentDiagram() == diag;
+  }
+
+  public AbstractDNode getMappingMSMStateMode(ECrossReferenceAdapter croffRef, AbstractState state, DDiagram diagram) {
+    final String mappingName;
```
pdulth, line 404: looks to be DiagramServices.getDiagramElement(diagram, state).getActualMapping

pdulth (changes requested):

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/ModeStateMachineServices.java
nlepine: Yes I am rewritting this methods.

core/plugins/org.polarsys.capella.core.sirius.analysis/description/common.odesign
nlepine: The filters preconditions were updated in the first patch but not in the second. I don't want to change this behavior because I don't reproduce the tools hidding. I am wrong ?

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/ModeStateMachineServices.java
pdulth: Create some states with an internal transition
image

Mark diagram as unsynchronized
image

Delete inner modes from diagram, 
- Using the tool "Show/Hide StateMode" and moving left both sub modes in the wizard.
image
- or you can use the "one-red-cross" button in the toolbar (it has the same effect)
image

Refresh, the computed link shall appear, even if the sub DNodes and DEdges no longer exists in the diagram (sourceRep /target are null)

nlepine: I have tried the same use case with LAB and component exchanges. During the refresh, the computed exchanges don't appear with an unsynchronised diagram, only with a synchronised diagram. Could you confirm this behavior ?
When I look at the DDiagramSychronizer, DEdgeCandidate are only created if the diagram is Synchronized : org.eclipse.sirius.diagram.business.internal.sync.DDiagramSynchronizer.refreshEdgeMapping(DiagramMappingsManager, Map<DiagramElementMapping, Collection<EdgeTarget>>, EdgeMapping, Map<EdgeMapping, Collection<MappingBasedDecoration>>, Map<String, Collection<SemanticBasedDecoration>>, IProgressMonitor)
image
Is there something I'm missing ?

core/plugins/org.polarsys.capella.core.data.migration/src/org/polarsys/capella/core/data/migration/contribution/ActivateDiagramFiltersContribution.java
```diff
@@ -0,0 +1,69 @@
...
+public class ActivateDiagramFiltersContribution extends AbstractMigrationContribution {
+
+  @Override
+  public void unaryEndMigrationExecute(ExecutionManager executionManager, Resource resource, MigrationContext context) {
+    super.unaryEndMigrationExecute(executionManager, resource, context);
+    activateRelevantDiagramFilters(resource);
+  }
+
+  private void activateRelevantDiagramFilters(Resource resource) {
+    for (DDiagram diagram : getAllMSMDiagrams(resource)) {
```
pdulth, line 41: can you enable activation of filter only if model is not in 6 version ?
(if major < 6)

(something like https://github.com/eclipse/capella/blob/409a90c4a2ed6c9c355badee0d068c32ec47f195/core/plugins/org.polarsys.capella.core.data.migration/src/org/polarsys/capella/core/data/migration/contribution/EnableFCAndPPFiltersMigrationContribution.java#L47-L62 )

pdulth (changes requested):

core/plugins/org.polarsys.capella.core.data.migration/plugin.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <?eclipse version="3.2"?>
 <!--
-  Copyright (c) 2006, 2021 THALES GLOBAL SERVICES.
+  Copyright (c) 2006, 2022 THALES GLOBAL SERVICES.
```
mPorhel, line 4: Copyright must be changed before merge.
Copyright (c) 2006, 2022 THALES GLOBAL SERVICES and others.

core/plugins/org.polarsys.capella.core.data.migration/src/org/polarsys/capella/core/data/migration/contribution/ActivateDiagramFiltersContribution.java
```diff
@@ -0,0 +1,69 @@
+/*******************************************************************************
+ * Copyright (c) 2022 THALES GLOBAL SERVICES.
```
mPorhel, line 2: Copyright must be changed before merge.

Copyright (c) 2022 Obeo

* Contributors:
 *    Obeo - initial API and implementation
 
 (or nlepin, or funder of the change, + bug ID or intitial implementation)

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/ModeStateMachineServices.java
```diff
@@ -12,9 +12,13 @@
  *******************************************************************************/
 package org.polarsys.capella.core.sirius.analysis;
```
mPorhel, line 13: copyright should be updated

mPorhel (changes requested):
@nlepine  could you check all copyright sections for modified and added files ?

pdulth:
for me looks good. @mPorhel ?

PR #2311, opened 2022-01-17
#2312 fixing the error that occurs when creating a new Interaction

Iulius3456:
Change-Id: I0d58c60ffdd7d2c39dc2f3868e9145c9d2bf82f1
Signed-off-by: Cazoni Iulius-Alexandru <iulius.cazoni@thalesgroup.com>

### DIFF
--- a/core/plugins/org.polarsys.capella.core.model.helpers/src/org/polarsys/capella/core/model/helpers/ComponentExt.java
+++ b/core/plugins/org.polarsys.capella.core.model.helpers/src/org/polarsys/capella/core/model/helpers/ComponentExt.java
@@ -1733,8 +1733,10 @@ public class ComponentExt {
    */
   public static List<Part> getSubParts(Component component) {
     List<Part> result = new ArrayList<Part>();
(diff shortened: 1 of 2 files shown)


core/plugins/org.polarsys.capella.core.model.helpers/src/org/polarsys/capella/core/model/helpers/ComponentExt.java
```diff
@@ -1743,9 +1745,11 @@ public static ComponentArchitecture getRootComponentArchitecture(Component compo
...
-      for (Part part : component.getRepresentingParts()) {
-        result.addAll(PartExt.getDeployedParts(part));
+    if (component != null) {
+      if (useDeploymentLinks) {
+        for (Part part : component.getRepresentingParts()) {
+          result.addAll(PartExt.getDeployedParts(part));
+        }
       }
     }
     result.addAll(getSubParts(component));
```
pdulth, line 1755: Can you move this line inside the `if` too? It will avoid creation of a new empty ArrayList and an addAll

PR #2336, opened 2022-02-08
Master 6.0 workspace in use and perfo

arnauddieumegard:


arnauddieumegard:
This performance improvement does not improves anything.

Issue #2396, opened 2022-05-17
[story image] The rich text editor can be very long to display the content


lfasani:
Steps to reproduce
* open the attached capella project from [testCapellaWithLargeDescription.zip](https://github.com/eclipse/capella/files/8707378/testCapellaWithLargeDescription.zip)
* open the Properties view
* select "Root Operational Activity"
* select the description tab
-> it is very long to display

pdulth:
After some tests, it seems it breaks the wizard edition

- Add a description on an element
- Double click on the element in project explore
- Edit the description through the wizard, save, ok
- the Property view still keep the previous text.

![Fix](https://user-images.githubusercontent.com/34767327/169032830-0f92cb08-2a6b-4cf5-95ae-adcc02eab249.gif)

PR #2397, opened 2022-05-17
#2396 Avoid the description tab to be refreshed twice

lfasani:
Waiting 100ms before scheduling the description tab is not sufficient.
So it is now refreshed immediately if the element has changed only.

### DIFF
--- a/core/plugins/org.polarsys.capella.core.ui.properties.richtext/src/org/polarsys/capella/core/ui/properties/richtext/sections/CapellaDescriptionPropertySection.java
+++ b/core/plugins/org.polarsys.capella.core.ui.properties.richtext/src/org/polarsys/capella/core/ui/properties/richtext/sections/CapellaDescriptionPropertySection.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017, 2020 THALES GLOBAL SERVICES.
+ * Copyright (c) 2017, 2022 THALES GLOBAL SERVICES.
  * 
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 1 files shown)


core/plugins/org.polarsys.capella.core.ui.properties.richtext/src/org/polarsys/capella/core/ui/properties/richtext/sections/CapellaDescriptionPropertySection.java
```diff
@@ -120,11 +120,13 @@ public IStatus runInUIThread(IProgressMonitor monitor) {
    */
   @Override
   public void loadData(EObject descriptorOrCapellaElement) {
-    super.loadData(descriptorOrCapellaElement);
-    mapDescriptionSectionToEObject.put(CapellaDescriptionPropertySection.this, descriptorOrCapellaElement);
+    if (descriptorOrCapellaElement != null && !descriptorOrCapellaElement.equals(job.current)) {
+      super.loadData(descriptorOrCapellaElement);
```
pdulth, line 124: I think the super.loadData shall be outside the if.
https://github.com/eclipse/capella/commit/c52365a209b226c4f3888e77aae32a841601f93f#diff-0481f92d1e0406a7269ab3ad3a5042d0018ead747aa11f18c97b231dd193a890R272

pdulth:
After some tests, it seems it breaks the wizard edition

- Add a description on an element
- Double click on the element in project explore
- Edit the description through the wizard, save, ok
- the Property view still keep the previous text.

Fix

Issue #2416, opened 2022-06-23
 Due to Actor migration EPIC, table on SA "System Actors - Operational Actors/Operational Entities" does not work


Iulius3456:
X is not added to OA 1 - OA 1 position, although OA 1 at SA lvl has realized OA 1 at OA level.

![image](https://user-images.githubusercontent.com/87373904/175269939-95a42fbd-c885-4cfb-b040-2123ba40ba1b.png)

PR #2418, opened 2022-06-23
#2416 fixed Operational SA - Operational Actors/Entities table.

Iulius3456:


Iulius3456:
the fix isn't for this version

PR #2440, opened 2022-08-03
#1559 branch add goto menu for catalog element link

achour94:
review-requested


pdulth:
Will be fixed with a more general

Issue #2514, opened 2022-11-26
Update Library property tester and Library manager and provider APIs to use current shared project to scan for library in remote repository


SteveMonnier:
To bypass the limitation of library management for shared capella project (add/remove reference to shared library project), a property tester requires to be updated and an API added:
* CapellaModelIsLocalPropertyTester now uses the airdFile to check that the session is opened and contains a Capella model
* ILibraryManager and ILibraryProvider methods getAvailableModels has been doubled to give the session has a parameter. This allows CDOLibraryProvider to use the session to check for other models on the CDO repository it is connected to.

PR #2515, opened 2022-11-26
#2514 Update library property tester and provider/manager APis

SteveMonnier:
To bypass the limitation of library management for shared capella project (add/remove reference to shared library project), a property tester requires to be updated and an API added:
- CapellaModelIsLocalPropertyTester now uses the airdFile to check that the session is opened and contains a Capella model
- ILibraryManager and ILibraryProvider methods getAvailableModels has been doubled to give the session has a parameter. This allows CDOLibraryProvider to use the session to check for other models on the CDO repository it is connected to.

Change-Id: If79cedfe0128cc960f3da679706c5be85557c1e2
Signed-off-by: Steve Monnier <steve.monnier@obeo.fr>

### DIFF
--- a/common/plugins/org.polarsys.capella.common.libraries/src/org/polarsys/capella/common/libraries/ILibraryManager.java
+++ b/common/plugins/org.polarsys.capella.common.libraries/src/org/polarsys/capella/common/libraries/ILibraryManager.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2006, 2020 THALES GLOBAL SERVICES.
+ * Copyright (c) 2006, 2022 THALES GLOBAL SERVICES.
  * 
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -31,7 +31,13 @@ public abstract class ILibraryManager {
   /**
    * Returns all available models
    */
+  @Deprecated
   public abstract Collection<IModelIdentifier> getAvailableModels();
+
+  /**
+   * Returns all available models for the given {@link TransactionalEditingDomain}
+   */
+  public abstract Collection<IModelIdentifier> getAvailableModels(TransactionalEditingDomain domain);
 
   /** 
    * Returns all models related to the given editing domain.
--- a/common/plugins/org.polarsys.capella.common.libraries/src/org/polarsys/capella/common/libraries/manager/LibraryManager.java
+++ b/common/plugins/org.polarsys.capella.common.libraries/src/org/polarsys/capella/common/libraries/manager/LibraryManager.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2006, 2020 THALES GLOBAL SERVICES.
+ * Copyright (c) 2006, 2022 THALES GLOBAL SERVICES.
  * 
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -42,11 +42,6 @@ public class LibraryManager extends ILibraryManager implements ILibraryProviderL
   Collection<ILibraryProvider> providers = null;
 
   /**
-   * A stored version of all model identifiers. it is cleared each time a provider notifies us.
-   */
-  Collection<IModelIdentifier> models = null;
-
-  /**
    * A stored version of model per editing domain.
    */
   HashMap<TransactionalEditingDomain, IModel> modelsPerDomain = null;
@@ -77,23 +72,25 @@ public class LibraryManager extends ILibraryManager implements ILibraryProviderL
 
   @Override
   public Collection<IModelIdentifier> getAvailableModels() {
-    if (models == null) {
-      models = new ArrayList<IModelIdentifier>();
-      for (ILibraryProvider provider : getProviders()) {
-        for (IModelIdentifier model : provider.getAvailableModels()) {
-          if (!models.contains(model)) {
-            models.add(model);
-          }
-        }
-      }
-    }
-    return models;
+      return getAvailableModels(null);
   }
+
+  @Override
+  public Collection<IModelIdentifier> getAvailableModels(TransactionalEditingDomain domain) {
+      Collection<IModelIdentifier> models = new ArrayList<IModelIdentifier>();
+      for (ILibraryProvider provider : getProviders()) {
+          for (IModelIdentifier model : provider.getAvailableModels(domain)) {
+              if (!models.contains(model)) {
+                  models.add(model);
+              }
+          }
+      }
+      return models;
+    }
 
   @Override
   public void onLibraryProviderChanged(LibraryProviderEvent event) {
     // We should clear cache (maybe we could do less)
-    models = null;
     modelsPerDomain = null;
   }
 
@@ -101,7 +98,7 @@ public class LibraryManager extends ILibraryManager implements ILibraryProviderL
   public Collection<IModel> getAllModels(TransactionalEditingDomain domain) {
     HashMap<IModelIdentifier, IModel> models = new HashMap<IModelIdentifier, IModel>();
 
-    for (IModelIdentifier identifier : getAvailableModels()) {
+    for (IModelIdentifier identifier : getAvailableModels(domain)) {
       IModel model = getModel(domain, identifier);
       if (model != null) {
         models.put(identifier, model);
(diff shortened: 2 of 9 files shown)


core/plugins/org.polarsys.capella.core.libraries.ui/src/org/polarsys/capella/core/libraries/ui/propertyTester/CapellaModelIsLocalPropertyTester.java
```diff
@@ -33,7 +35,12 @@ public boolean test(Object receiver_p, String property_p, Object[] args_p, Objec
       // the project is exported if there is no semantic model file at the same level of the aird file
       if ((selection != null) && (selection instanceof IFile)) {
         IFile airdFile = (IFile) selection;
-        return new IFileRequestor().search(airdFile.getParent(), CapellaResourceHelper.CAPELLA_MODEL_FILE_EXTENSION, false).size() > 0;
+        Session session = SessionManager.INSTANCE.getSession(URI.createURI(airdFile.getLocationURI().toString()), new NullProgressMonitor());
```
pdulth, line 38: The property tester shall be removed as the menu is now displayed in both cases, local and remote. 
Moreover, the getSession(uri) call will force to load the sirius session if we right click on the capella aird, even if it is closed. Session will be not opened but created (ie. loaded)
https://github.com/pcdavid/org.eclipse.sirius/blob/aee60e43454d04ff568b9ebd4e12f47d7661f19e/plugins/org.eclipse.sirius/src/org/eclipse/sirius/business/internal/session/SessionManagerImpl.java#L264-L282

common/plugins/org.polarsys.capella.common.libraries/src/org/polarsys/capella/common/libraries/ILibraryManager.java
```diff
@@ -10,53 +10,59 @@
...
+public abstract class ILibraryManager {
+
+  public static ILibraryManager INSTANCE = new LibraryManager();
+
+  public static TransactionalEditingDomain DEFAULT_EDITING_DOMAIN = null;
+
+  /**
+   * Returns all available models
+   */
+  public abstract Collection<IModelIdentifier> getAvailableModels();
```
pdulth, line 35: No longer used, to set deprecated.

common/plugins/org.polarsys.capella.common.libraries/src/org/polarsys/capella/common/libraries/ILibraryManager.java
```diff
@@ -10,53 +10,59 @@
...
+
+  /**
+   * Returns all available models for the given {@link Session}
+   */
+  public abstract Collection<IModelIdentifier> getAvailableModels(Session session);
```
pdulth, line 40: TransactionalEditingDomain domain rather than session

common/plugins/org.polarsys.capella.common.libraries/src/org/polarsys/capella/common/libraries/manager/LibraryManager.java
```diff
@@ -100,8 +112,14 @@ public void onLibraryProviderChanged(LibraryProviderEvent event) {
...
     HashMap<IModelIdentifier, IModel> models = new HashMap<IModelIdentifier, IModel>();
+    List<AirdResource> airdResources = domain.getResourceSet().getResources().stream().filter(AirdResource.class::isInstance).map(AirdResource.class::cast).collect(Collectors.toList());
+    Iterator<AirdResource> airdResourcesIterator = airdResources.iterator();
+    Session session = null;
+    while (session == null && airdResourcesIterator.hasNext()) {
+        session = SessionManager.INSTANCE.getSession(airdResourcesIterator.next().getURI(), new NullProgressMonitor());
+    }
 
-    for (IModelIdentifier identifier : getAvailableModels()) {
+    for (IModelIdentifier identifier : getAvailableModels(session)) {
```
pdulth, line 122: domain

common/plugins/org.polarsys.capella.common.libraries/src/org/polarsys/capella/common/libraries/manager/LibraryManager.java
```diff
@@ -100,8 +112,14 @@ public void onLibraryProviderChanged(LibraryProviderEvent event) {
   @Override
   public Collection<IModel> getAllModels(TransactionalEditingDomain domain) {
     HashMap<IModelIdentifier, IModel> models = new HashMap<IModelIdentifier, IModel>();
+    List<AirdResource> airdResources = domain.getResourceSet().getResources().stream().filter(AirdResource.class::isInstance).map(AirdResource.class::cast).collect(Collectors.toList());
```
pdulth, line 115: the session lookup can be moved to CDOLibraryProvider. 

this code assumes that there is an aird and a session in the current domain, which may not be the case in this api

common/plugins/org.polarsys.capella.common.libraries/src/org/polarsys/capella/common/libraries/provider/ILibraryProvider.java
```diff
@@ -32,6 +33,11 @@ public interface ILibraryProvider {
    */
   public Collection<IModelIdentifier> getAvailableModels();
 
+  /**
+   * Returns all models available for the given {@link Session} through the provider
+   */
+  public Collection<IModelIdentifier> getAvailableModels(Session session);
```
pdulth, line 39: domain

pdulth (changes requested):

core/plugins/org.polarsys.capella.core.libraries.ui/src/org/polarsys/capella/core/libraries/ui/propertyTester/CapellaModelIsLocalPropertyTester.java
SteveMonnier: Oh right, I should have used getExistingSession here. 
I kept the property tester as a way to be sure that it only enable the menu if it is a project with a capella semantic model. Note that if I remove the property tester, for a modeling project without a capella semantic model, the menu is enabled but clicking on it does not open the Library dialog, so it does not lead to an issue, but it looks better to keep this property tester.

common/plugins/org.polarsys.capella.common.libraries/src/org/polarsys/capella/common/libraries/ILibraryManager.java
SteveMonnier: done

common/plugins/org.polarsys.capella.common.libraries/src/org/polarsys/capella/common/libraries/manager/LibraryManager.java
SteveMonnier: I used session instead of domain here because in com.thalesgroup.mde.melody.collab.ui.provider.CDOLibraryProvider.getAvailableModels(Session) the session is used twice but there is no api for the domain. I could find the session from the domain though. Did you have a specific reason to want this new API to be with domain instead of session?

common/plugins/org.polarsys.capella.common.libraries/src/org/polarsys/capella/common/libraries/manager/LibraryManager.java
SteveMonnier: The session handling here was just done so it could forward it to CDOLibraryProvider. Ok I get why you wanted a previous api with domain instead of session. I am updating it.

common/plugins/org.polarsys.capella.common.libraries/src/org/polarsys/capella/common/libraries/manager/LibraryManager.java
SteveMonnier: I have my answer from the next comment.
Fixed

common/plugins/org.polarsys.capella.common.libraries/src/org/polarsys/capella/common/libraries/provider/ILibraryProvider.java
SteveMonnier: done

SteveMonnier:
I pushed a new version that only adds the note about libraries and viewpoints in the documentation

pdulth (approved):

Issue #2522, opened 2022-12-12
Default Perspective : replace FastLinker capability with Outline 


etraisnel2:
Once outline view is completed and we are ensured that it doesn't cause any performance issue, 
The outline view shall replace fastlinker in the default perspective 

Capella shall also contributes to the outline menu (see https://bugs.eclipse.org/bugs/show_bug.cgi?id=581205)

all documentation and tests should be updated

PR #2523, opened 2022-12-13
#2522 Switch Fast linker and outline view in default perspective

etraisnel2:
Temporary commit : need refinement and doc update + tests -> for performance test purpose

Change-Id: I986c617b96334d399b9b75a0110c765193a94588


core/plugins/org.polarsys.capella.core.sirius.ui/plugin.xml
```diff
@@ -204,7 +204,22 @@
             </visibleWhen>
          </command>
       </menuContribution>
-         
+       <menuContribution
```
pdulth, line 207: Can you move it to the other menu group contributions 
(in https://github.com/eclipse/capella/blob/master/core/plugins/org.polarsys.capella.core.platform.sirius.ui.menu/plugin.xml#L31
)

pdulth (changes requested):

pdulth (approved):

Issue #2526, opened 2023-01-04
File extension needed in filename when exporting metrics


etraisnel2:
in Capella 6, when exporting Capella metrics, or Progress Monitoring
the export dialog requires a filename that contains the extension or the export fails

this behaviour also appear in all inheriting classes of AbstractExportDialog, like in filtering (https://github.com/eclipse/capella-filtering/issues/74)

PR #2528, opened 2023-01-04
#2526 Export Dialog / Export validation preference filename fix

etraisnel2:
If extension was missing from filename, the export would always fail Now ensures that the filename ends with the file extension, or add it if missing

Change-Id: I47197c67a1039b28b6b0fd94e9550eac69320c13


common/plugins/org.polarsys.capella.common.ui/src/org/polarsys/capella/common/ui/toolkit/dialogs/AbstractExportDialog.java
```diff
@@ -108,6 +108,18 @@ protected void exportPressed() {
     //
     String fileName = fd.open();
 
+    
+    int filterIndex = fd.getFilterIndex();
+    if(filterIndex != -1) {
+      //If a filter was selected
+      String selectedFilter = fd.getFilterExtensions()[filterIndex];
+      String fileExtension = selectedFilter.substring(1);
+      if(! fileName.endsWith(fileExtension)) {
```
pdulth, line 117: can you use capella.epf file formatting

pdulth (approved):

Issue #2534, opened 2023-01-05
Copy-paste layout in FCD and PPD not working properly


etraisnel2:
When 2 FC (or PP) reference the same function on a diagram, copy pasting the layout of that function from another diagram to this diagram ends up piling up the two instances of that function

PR #2535, opened 2023-01-05
#2534 Fixed copy paste layout for PPD and FCD

etraisnel2:
When 2 FC (or PP) reference the same function on a diagram, copy pasting the layout of that function from another diagram to this diagram, ended up piling up the two instances of that function

Fixed it

Change-Id: I4dac88b331b8cb9d8058ac3f1466ecf2b588d32f


etraisnel2:
Unit tests have not been pushed yet

pdulth (approved):

Issue #2554, opened 2023-01-24
Fix Expert Properties Capability + Expert and Semantic Properties shall be hidden by default


etraisnel2:
Expert and Semantic Properties shall be displayed by defult 
the Capability "Expert properties" is supposed to do so but it doesn't properly work

-> Fix this capability
-> Disable this capability by default 

Since these properties views are often used to get the ID , create a new action "Copy Unique Identifier"

Modify "Copy as Description" to "Copy as HyperLink For Description", to be more explicit for the user.

Fix "Copy as Qualified Name" to handle multi-selection

Fix "Copy as Text" on Representations , so that their named is properly copied

"Copy As text" on a category in semantic browser shall work

Reorder "Copy as" menu items to appear in this order :
- as Text
- as Qualified Name
- as Hyperlink for description
- as Unique Identifier

Disable FastLinker capability by default

Modify the embedded doc so that any mentions of it are uptodate

PR #2565, opened 2023-02-01
#2554 Filter property sections via activity bindings

pdulth:
Property sections can now be filtered by activities, bind it to an activity that matches 'sectionId'.

Tabs are hidden by default if there is no sections in it

Automatic refresh when activities change is not performed. The user has to re-select the element to make tab changes visible in the view.


pdulth:
Missed cherry pick again :D

irina-toma (approved):

Issue #2570, opened 2023-02-01
Add "Viewpoint Manager" in Windows > Show view


etraisnel2:
Add Viewpoint Manager in view shortcut in Capella Perspective

PR #2571, opened 2023-02-01
#2570 Added Viewpoint Manager view in "Show view" menu

etraisnel2:
Modified Perspective to include this shortcut

Change-Id: I53dd1615d02770ff1d2ea857b86646e65f7fee70


pdulth (changes requested):
Can you add it on the plugin.xml instead
https://github.com/eclipse/capella/blob/77e6f88dbe1c8534436bfc245bfa385e5268f7b2/core/plugins/org.polarsys.capella.core.platform.sirius.ui.perspective/plugin.xml#L31

pdulth (approved):

Issue #2596, opened 2023-02-23
Double-clicking on Capella Elements in SB doesn' t open the properties wizard


etraisnel2:
Regression from #2452 and #2449 

Properties wizard cannot be opened anymore in Semantic browser on pressing enter or double clicking

Fix it

Realign how multi-selection works in Capella Project Explorer so that both SB an PJ works the same way

Add JUnit Test

PR #2597, opened 2023-02-23
#2596 Properties wizard can now be opened on double click in semantic

etraisnel2:
browser

Fixed issue where double-clicking wouldn't open properties wizard anymore

Fixed multi-selection in both project explorer and SB When pressing enter, the corresponding action is ran for each selected elements
actions being :
open diagram on representation descriptor
navigate on chain (FC/PP/OP) (create if 0, open if 1, ask if more than 1 diagram)

Updated condtions of ModelElementActionProvider so that it can also take representations descriptors

Change-Id: I4a2052c23ac73d218aa911a78842de99826e6d2f

### DIFF
--- a/core/plugins/org.polarsys.capella.core.platform.sirius.ui.navigator/plugin.xml
+++ b/core/plugins/org.polarsys.capella.core.platform.sirius.ui.navigator/plugin.xml
@@ -480,9 +480,18 @@
                class="org.polarsys.capella.core.platform.sirius.ui.navigator.actions.providers.ModelElementActionProvider"
                id="capella.project.explorer.modelElementActionProvider">
             <enablement>
-               <instanceof
-                     value="org.polarsys.kitalpha.emde.model.Element">
-               </instanceof>
+               <or>
+                  <instanceof
+                        value="org.polarsys.kitalpha.emde.model.Element">
+                  </instanceof>
+                  <instanceof
+                        value="org.eclipse.sirius.viewpoint.DRepresentationDescriptor">
+                  </instanceof>
+                   <test
+                        forcePluginActivation="false"
+                        property="org.polarsys.capella.core.platform.sirius.ui.navigator.itemWrapperPropertyTester">
+                  </test>
+              </or>
             </enablement>
          </actionProvider>
           <dropAssistant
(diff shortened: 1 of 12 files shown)


tests/plugins/org.polarsys.capella.test.semantic.ui.ju/src/org/polarsys/capella/test/semantic/ui/ju/testcases/SemanticBrowserReferencingElementNavigationTest.java
```diff
@@ -147,7 +151,76 @@ public void test() throws Exception {
 
       IEditorPart editor = EclipseUIUtil.getActiveEditor();
       assertTrue("We should have a DDiagramEditor", editor instanceof DDiagramEditor);
+      
+      //Test that double-clicking an element (Capability here) opends a property wizard
```
minhtutonthat, line 155: Here we're trying to do an integration test. Would it be better to do some Unit tests instead? 
For example, an idea would be:
- Separate the logic that handles the double-click into handler classes (e.g. a generic interface ISemanticBrowserDoubleClickingHandler and 2 classes inherting that interface ElementDoubleClickingHandler and DiagramDoubleClickingHandler)
- Add a new doubleClickingHandler field to the SemanticBroswerView class
- Change the handleDoubleClick method to associate the doubleClickingHandler field to the correct handler depending on clicked element and call the handler eventually.
- Now in the test we can just test that the handler is the correct instance depending on what's selected.
- We can go further to unit test the handlers to see if they really open the wizard and the diagram but I think it's not the scope of this test.

etraisnel2: I have tried to follow your approach but couldn't properly extract the different behavior into specific handlers
Since on a multiple selection we iterate over each element, the logic may differ .
Instead, to be able to do unit tests, I used boolean methods to know what can action is achieved when double-clicking a given element

minhtutonthat (approved):

Issue #2634, opened 2023-03-31
Optimize Capella validation


lfasani:
A lot of time is spent iterating on the model during the Validation of the Capella model:

* DWF_I_23 takes 49% og total time of OpSystem (not provided with this ticket)
![image](https://user-images.githubusercontent.com/12660743/229113213-f5805c9c-72bb-4d46-9a0b-74193de68b41.png)

Using a cache for `GetAllQueries.getAll ` it could be possible to optimize the time spent
Maybe using aql could be a good thing because AQL leverage the meta-model information to prune when iterating over the model

* MD_CHK_StateTransition takes 16%

PR #2641, opened 2023-04-12
#2634 Optimization of ModelCache

etraisnel2:
Moved ModelCache to common packages

Added support to cache function with 2 parameters

Added support for cachedfunction and cachedbifunction, so that methods that retrieve from cache can be directly aclled on it (through .get)

Cache is now used for getAllQueries and GetAvailable_StateTransitionTrigger which both caused performance loss because cache wasn't used

ModelCache.getCache is now set to deprecated, ModelCache.getOrApply should be used instead

Change-Id: Ieaf149bc991cd9bc1bd5efd25560270d7fa868fc


common/plugins/org.polarsys.capella.common/src/org/polarsys/capella/common/helpers/Triplet.java
```diff
@@ -0,0 +1,89 @@
+package org.polarsys.capella.common.helpers;
```
pdulth, line 1: To delete, this class is defined in org.polarsys.capella.common.helpers.cache
Getfirst methods shall be moved from here

common/plugins/org.polarsys.capella.common.helpers/src/org/polarsys/capella/common/helpers/cache/Cache.java
```diff
@@ -0,0 +1,162 @@
...
+import java.util.stream.Collectors;
+
+import org.polarsys.capella.common.helpers.Triplet;
+
+public class Cache {
+
+  private Map<Object, Object> cachedResult = new ConcurrentHashMap<>();
+
+  @SuppressWarnings("unchecked")
+  @Deprecated
```
pdulth, line 32: As a CachedFunction is a Function, we can keep this one and delete the next get(CachedFunction) one

common/plugins/org.polarsys.capella.common.helpers/src/org/polarsys/capella/common/helpers/cache/ModelCache.java
```diff
@@ -32,6 +33,7 @@ private ModelCache() {
    * @return If enabled, return the cached result if any or apply the function to the given parameter and cache the
    *         result before returning it.
    */
+  @Deprecated
```
pdulth, line 36: can you add a comment to explain why it's deprecated. 
(Use a CachedFunction instead to ensure that the cache is shared with all calls to this function rather than having a cache per calls )

pdulth (changes requested):

common/plugins/org.polarsys.capella.common.helpers/src/org/polarsys/capella/common/helpers/EObjectExt.java
```diff
@@ -38,16 +41,38 @@
...
+      .getAllQueries()::getAll;
+
   /**
    * @param currentElement
    * @param targetType
    * @return Set<EObject>
    */
   public static Set<EObject> getAll(EObject currentElement, EClass targetType) {
-    List<EClass> filters = new ArrayList<EClass>();
+    return new HashSet<EObject>(getAll_withEClass.get(currentElement, targetType));
```
pdulth, line 55: LinkedHashSet to keep order as defined in the original method. it was probably important

common/plugins/org.polarsys.capella.common.helpers/src/org/polarsys/capella/common/helpers/EObjectExt.java
```diff
@@ -38,16 +41,38 @@
...
     return MDEQueries.getInstance().getAllQueries().getAllFiltered(currentElement, targetType, filters);
   }
 
+  /**
+   * @param currentElement
+   * @param targetType
+   * @return Set<EObject>
+   */
+  public static Set<EObject> getAll(EObject currentElement, Predicate<EObject> predicate) {
+    return new HashSet<EObject>(getAll_withPredicate.get(currentElement, predicate));
```
pdulth, line 73: same here

pdulth (changes requested):

common/plugins/org.polarsys.capella.common.helpers/src/org/polarsys/capella/common/helpers/cache/Cache.java
etraisnel2: Done

common/plugins/org.polarsys.capella.common.helpers/src/org/polarsys/capella/common/helpers/EObjectExt.java
etraisnel2: Done

common/plugins/org.polarsys.capella.common/src/org/polarsys/capella/common/helpers/Triplet.java
etraisnel2: Done

pdulth (approved):

PR #2646, opened 2023-04-20
[releng] Bump Capella to 7.0.0

etraisnel2:
Change-Id: I98cbb0b546660a2ef3ca3d1980c8088711644128


etraisnel2:
M2 has not been bumped yet
test model not migrated
only pom and manifest

PR #2665, opened 2023-06-23
[releng] Allow migration from Major version 6 to 7

etraisnel2:
Change-Id: I3e78602ba8457a2c5b99a34c8b3f9be7f1a7fc08


core/plugins/org.polarsys.capella.core.data.migration/src/org/polarsys/capella/core/data/migration/af/ViewpointMigrationContribution.java
```diff
@@ -204,8 +207,15 @@ private boolean isMigrationPossible(Version fileVersion, Version currentVersion)
...
-    
-    // Allow migration from 1.4.x to 6.x and from 5.x to 6.x
-    return (currentVersion.getMajor() == 6) && (fileVersion.getMajor() == 5 || (fileVersion.getMajor() == 1 && fileVersion.getMinor() == 4));
+    return (allowedMigrations.get(fileVersion.getMajor()) == currentVersion.getMajor());
+  }
+
+  private static HashMap<Integer, Integer> initializeMigrationMap() {
+    HashMap<Integer, Integer> migrationMap = new HashMap<Integer, Integer>();
+    // Allow migration from 1 to 4, from 5 to 6 and from 6 to 7
+    migrationMap.put(1, 4);
```
pdulth, line 216: I think it's not ok, 
We are on master/7.x, we allow migration towards 7.x (allowing to 4 or 5 will never occurs ;)
1.4 to 7
5.x to 7
6.x to 7
7.x to 7

pdulth (changes requested):

core/plugins/org.polarsys.capella.core.data.migration/src/org/polarsys/capella/core/data/migration/af/ViewpointMigrationContribution.java
pdulth: Maybe you can do somthing like : 
if (fileVersion.getMajor == 0)
  return false;
if (fileVersion.getMajor == 1)
  return fileVersion.getMinor == 4;
if (fileVersion.getMajor <  currentVersion.getMajor())
  return true;
if (fileVersion.getMajor ==  currentVersion.getMajor())
  return fileVersion.getMinor <=  currentVersion.getMinor();
return false;

it will allow any kind of migration towards current version. I will upadte the wiki to ensure next bump will look at this file if its specific case.

pdulth (approved):

PR #2682, opened 2023-07-25
2681 [Capella properties] Bad concurrent modification

SteveMonnier:
Refresh Capella widgets when they are enabled in case its semantic target was updated remotely.

### DIFF
--- a/core/plugins/org.polarsys.capella.core.data.core.properties/src/org/polarsys/capella/core/data/core/properties/fields/BasicElementGroup.java
+++ b/core/plugins/org.polarsys.capella.core.data.core.properties/src/org/polarsys/capella/core/data/core/properties/fields/BasicElementGroup.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2006, 2020 THALES GLOBAL SERVICES.
+ * Copyright (c) 2006, 2023 THALES GLOBAL SERVICES.
  * 
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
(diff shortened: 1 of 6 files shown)


core/plugins/org.polarsys.capella.core.data.core.properties/src/org/polarsys/capella/core/data/core/properties/fields/BasicElementGroup.java
```diff
@@ -145,5 +146,9 @@ public void enableSummaryField(boolean enabled) {
   public void setEnabled(boolean enabled) {
     LockHelper.getInstance().enable(nameTextField, enabled);
     LockHelper.getInstance().enable(summaryTextField, enabled);
+	if (enabled && semanticElement instanceof CapellaElement) {
```
pdulth, line 149: the BasicElementGroup is more wide than CapellaElement.

core/plugins/org.polarsys.capella.core.data.core.properties/src/org/polarsys/capella/core/data/core/properties/fields/BasicElementGroup.java
```diff
@@ -145,5 +146,9 @@ public void enableSummaryField(boolean enabled) {
...
+		// Refresh widgets from semantic element in case of remote update
+		loadData(semanticElement);
```
pdulth, line 151: The setEnabled/enableButton/enabledxxx methods goal is to enable or disable the properties ui elements, it must not call load method to modify the internal state of the view. 
It shall have somewhere in Team for Capella that is allowing to reload the property .
Maybe in com.thalesgroup.mde.melody.collab.ui.lock.listeners.CDOReadOnlySectionHandler, maybe in notifyIsReleased

pdulth (changes requested):

core/plugins/org.polarsys.capella.core.data.core.properties/src/org/polarsys/capella/core/data/core/properties/fields/BasicElementGroup.java
SteveMonnier: Hi Philippe,
I will look into it, I made the change on the widget level because I was missing a more global way to trigger a refresh after the enablement (as it was expected to be needed only on focus change). I may have missed something.

SteveMonnier:
I managed to move the fix to T4C. Some APIs could be added in IReadOnlyListener to prevent some castings, but for my case it was enough. Thanks Philippe.

Issue #2687, opened 2023-08-02
Extract Requirement from Capella


etraisnel2:
Requirement should be extracted from Capella M2 and made as a "basic" addon.

Extract Requirement.ecore from Capella M2 and any relationship to it
Extract Requirement from odesign
Methods that provides Req capabilities shall be moved to Req

Validationn Rule "I_11" (requirement id check" to be extracted to basic-vp

Any capella feature that cannot be extracted shall be noted down :
SemanticModelUtil doesn't provide req anymore
DiffMerge SkeletonMatchPolicy doesn't use req anymore
CapellaValidatorAdapter : same

see https://github.com/eclipse/capella-basic-vp/issues/71

PR #2722, opened 2023-09-27
#2687 Remove Requirement from documentation

etraisnel2:
Change-Id: Ibb0552c73d3dc74a43d7d052e2d6d2d66de14afd


pdulth:
For the failing junit, 
the css was removed from validationrules.html file. It's because you generate html using right click/wiki text menu rather than using the build.xml file located at the root of the documentation plugins to generate.

Issue #2708, opened 2023-08-25
Contribute custom traces to trace manager 


etraisnel2:
Using an extension point mechanism, external plugins should be able to contribute their own kind of trace 
Initial need : Basic Requirement , that was extracted from Capella, must be able to contribute its RequirementsTrace, not to lose any behavior

PR #2729, opened 2023-09-29
#2708 Trace contributions enabled on vp activation

etraisnel2:
Contribution is now only enabled when the viewpoint is activated on the current model

Change-Id: I62a14fa734a24aa445cb22e027ef73e03d8f5aac

### DIFF
--- a/core/plugins/org.polarsys.capella.core.platform.eclipse.ui.trace/src/org/polarsys/capella/core/platform/eclipse/capella/ui/trace/extension/TraceExtensionManager.java
+++ b/core/plugins/org.polarsys.capella.core.platform.eclipse.ui.trace/src/org/polarsys/capella/core/platform/eclipse/capella/ui/trace/extension/TraceExtensionManager.java
@@ -21,22 +21,25 @@ import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IConfigurationElement;
 import org.eclipse.core.runtime.IExtensionRegistry;
 import org.eclipse.core.runtime.Platform;
+import org.eclipse.emf.ecore.EObject;
+import org.eclipse.emf.ecore.resource.ResourceSet;
 import org.polarsys.capella.common.data.modellingcore.AbstractTrace;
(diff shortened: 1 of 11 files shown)


core/plugins/org.polarsys.capella.core.platform.eclipse.ui.trace/src/org/polarsys/capella/core/platform/eclipse/capella/ui/trace/extension/TraceExtensionManager.java
```diff
@@ -21,24 +22,35 @@
...
 
 public class TraceExtensionManager {
 
   public static TraceExtensionManager eINSTANCE = new TraceExtensionManager();
-  static String EXTENSION_POINT_ID = "org.polarsys.capella.core.platform.eclipse.capella.ui.trace.extension";
-  static String EXTENSION_POINT_CLASS = "traceExtension";
-
+  String EXTENSION_POINT_ID = "org.polarsys.capella.core.platform.eclipse.capella.ui.trace.extension";
+  String EXTENSION_POINT_CLASS = "traceExtension";
+  private HashMap<ITraceExtension, Set<ResourceSet>> activatedExtensions ;
```
pdulth, line 38: Storing the resourceset will lead to memory leaks. as when session is closed, it will be remain here.
Does ViewpointManager.getInstance(rSet).isActive(anExtension.getViewpointID()) in your inActive method is not enough?

pdulth (changes requested):

core/plugins/org.polarsys.capella.core.platform.eclipse.ui.trace/src/org/polarsys/capella/core/platform/eclipse/capella/ui/trace/extension/TraceExtensionManager.java
etraisnel2: Yup you're right, removed the cache and now directly use vpManager.isActive

pdulth (approved):

PR #2735, opened 2023-10-06
#2709 propose similar Cport to PL component exchange

mchoubby:
a rework because previous solution propose component port that not related to a CE


pdulth:
I'm not sure which one to merge, both PR are modifiyng the same files.

pdulth (approved):

PR #2736, opened 2023-10-10
#2705 log info from accelerator commands-rework

mchoubby:


pdulth (changes requested):
Can you fix formatting. There is many files that are unexpectly changed.
(maybe start over master branch and modify only the one or two that you modified

core/plugins/org.polarsys.capella.core.platform.sirius.ui.actions/src/org/polarsys/capella/core/platform/sirius/ui/commands/ExchangeItemAllocationOnPortsCommand.java
```diff
@@ -105,41 +104,37 @@ protected Collection<ModelElement> retrieveModelElements(ModelElement modelEleme
...
+      if (fe.getSource() instanceof FunctionOutputPort) {
+        return processPort((FunctionOutputPort) fe.getSource(), flows);
       }
-
-      if (fe.getTarget() != null) {
-        if (fe.getTarget() instanceof FunctionInputPort) {
-          processPort((FunctionInputPort) fe.getTarget(), flows);
-        }
+      if (fe.getTarget() instanceof FunctionInputPort) {
+        return processPort((FunctionInputPort) fe.getTarget(), flows);
```
pdulth, line 126: for an exchange, maybe do both no ? 
now it processPort only on the source, no longer on the target

pdulth (changes requested):

core/plugins/org.polarsys.capella.core.platform.sirius.ui.actions/src/org/polarsys/capella/core/platform/sirius/ui/commands/ExchangeItemAllocationOnPortsCommand.java
mchoubby: i removed the fe.getSource() != null because it is useless, instanceof will take care of checking if it null.
i didnt change the logic
same fot fe.getTarget()

pdulth: the **return** on the line 123 for the source, makes that it is no longer called for the target.

mchoubby: done

pdulth (approved):

Issue #2739, opened 2023-10-12
Validation rule that checks if a Component Exchange is assigned to multiple PhysicalLinks


etraisnel2:
Due to a difference of copy paste behaviour between the diagram editor and the project explorer, it is possible to have multiple PhysicalLinks allocating the same ComponentExchange 

when copy/pasting PL
Diagram copy paste will only copy paste the edge representing the PL 
Project explorer copy paste will copy the PL and its contents (which is a CE Allocation)

validation rule DWF_DC_45
a Component Exchange shall be allocated to only one Physical Link

Quickfix : prompt for a physical allocation to keep, delete the others

PR #2745, opened 2023-10-17
#2739 Validation rule that checks if a Component Exchange is assigned to multiple PhysicalLinks

etraisnel2:


pdulth:
can you fix junit.
it's because you generate validation documentation without the build.xml

pdulth (approved):

Issue #2763, opened 2023-11-13
Auto-size new created elements in Capella diagrams


GlennPlou:
Some nodes are too small and have their labels cut off when they are created or inserted. These nodes should be auto-sized when created, according to their label, as is done for containers.
![image](https://github.com/eclipse/capella/assets/33932970/62995943-6593-4ee4-875d-0ac3989ead9c)

PR #2764, opened 2023-11-13
#2763 Set Size Computation Expression to -1 to use auto-size on nodes

GlennPlou:
Auto-size mode is now enabled on nodes with square style.


core/plugins/org.polarsys.capella.core.sirius.analysis/description/oa.odesign
lfasani, line 1: github says 2,756 additions 2,756 deletions. I guess there is a problem of line ending.
Could you change so that we can see just the real changes

lfasani (approved):

mPorhel:
Updated PR : https://github.com/eclipse/capella/pull/2786
This one can be abandonned, behavior and specification of auto-sized nodes have been reworked in Sirius.

GlennPlou:
See PR https://github.com/eclipse/capella/pull/2786

Issue #2662, opened 2023-06-21
Replace Welcome view with a SWT Implementation


etraisnel2:
In 7.0 
Edge Browser doesn't display the Welcome Page properly (local html being seen as a security issue by edge)
To fix that, we need a SWT Implementation of the Welcome View, so that links are displayed the same way they used to be

etraisnel2:
Small issue spotted when running the product

WelcomeView cannot be casted to CustomizableIntroPart

etraisnel2:
There still seems to be an issue with dispose
![image](https://github.com/eclipse/capella/assets/109157813/1e56788c-2db4-4411-b751-dc258f1a1c96)

PR #2782, opened 2023-12-20
#2662 Avoid ClassCastException on WelcomeView

etraisnel2:
Change-Id: Ie228d2257fe5f8ea5c5eabf22af9b96401b6545a


doc/plugins/org.polarsys.capella.core.ui.intro/src/org/polarsys/capella/core/ui/intro/views/WelcomeView.java
```diff
@@ -398,5 +401,9 @@ public void dispose() {
 	@Override
 	public void saveState(IMemento memento) {
 	}
+	
+	public Composite getControl(){
```
pdulth, line 405: format

pdulth (changes requested):

pdulth (approved):

PR #2788, opened 2024-01-10
Aggregation pour arianegroup v6.1

SteveMonnier:


lredor:
After exchange with @SteveMonnier and @ebausson-obeo, this PR was wrongly pushed. So I close it

Issue #2789, opened 2024-01-12
Pie icons not displayed when exported, or on first diagram opening


etraisnel2:
When multiple Functional Chains or Physical Paths overlap on a single graphical edge, edge color changes and a pie Icon representing all overlapping colors is displayed at both ends of the edge.

These icons are transient and not stored in the model nor the aird.

In your project or workspace preferences,if the "refresh on opening" preferences is set to false (which are default values for projects on T4C), the icons won't appear on opening, but only after the diagram is refreshed 
In the same way, exporting representations won't properly work if the diagrams are not opened and refreshed 

In order to fix this issue, edges for FunctionalExchange and PhysicalLink (on xAB and dataflows) should reference (using the semanticElements field) all displayed involving FC or PP (any displayed FC or PP that involve the edge)

PR #2790, opened 2024-01-12
#2789 Pie icon properly exports

etraisnel2:


pdulth:
Added Functional Exchange on semantic elements, and not rely on HashSet to compute them, as order may change between sessions

pdulth (approved):

PR #2775, opened 2023-12-15
[releng] Update to stable of Sirius 7.4.x

lfasani:

PR #2793, opened 2024-01-19
#2763 Enable auto-size for nodes with square style, add migration contributor and fix existing tests

GlennPlou:


pdulth:
Included into #2775

PR #2801, opened 2024-02-06
Fix disable.nebula.richtext property.

jeffallen:
(a new commit to trigger a new build)


pdulth:
Integrated your (slightly modified) commit into https://github.com/eclipse/capella/pull/2807

PR #2814, opened 2024-02-16
[releng] Fix memory issue on build

pdulth:
Maven process doesn't need that much memory, as its more sub processes that will require it. (eclipse director that will create product in that current case)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -59,20 +59,18 @@ pipeline {
 	       		}
 	     	}
 	    }
-	    
-    	stage('Build and Package') {
-      		steps {
-      			script {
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -59,20 +59,16 @@ pipeline {
...
-							sh "mvn verify -Pfull ${sign}"
-						}
-					}
-      			}
-	     	}
-	    }
-	    
+
+		stage('Build and Package') {
+			steps {
```
AnitaMateiGabriel, line 64: i assume sonar is being entirely taken care of in its separate stage so no need for credentials?

AnitaMateiGabriel (approved):

Issue #2822, opened 2024-02-28
Richtext : Description field should not be saved when element is owned by a closed session


etraisnel2:
See https://github.com/eclipse/kitalpha/issues/788

Description section keeps track of the last selected element
If this elements is owned by a closed session, an error is raised

PR #2823, opened 2024-02-28
#2822 Fix for Richtext description

etraisnel2:
If the tracked element is owned by a closed session, do no save it

Change-Id: I0000000000000000000000000000000000000000


core/plugins/org.polarsys.capella.core.ui.properties.richtext/src/org/polarsys/capella/core/ui/properties/richtext/fields/ElementDescriptionGroup.java
```diff
@@ -212,6 +214,12 @@ public void aboutToBeShown() {
     if (updateDescriptionEditability(semanticElement, semanticFeature)) {
       try {
         if (semanticElement != null && semanticFeature != null) {
+          Session session = SessionManager.INSTANCE.getSession(semanticElement);
```
pdulth, line 217: in aboutToBeHidden no?

pdulth (approved):

PR #2824, opened 2024-02-28
#2700 PhysicalLink ends are now displayed in referenced panel

etraisnel2:
Since PhysicalLinks are not oriented, there can't be target nor source instead, just display both ends at the same time


pdulth:
@etraisnel2  Can you look, build is KO

pdulth:
https://ci.eclipse.org/capella/job/capella-product/view/change-requests/job/PR-2824/3/console

pdulth (approved):

PR #2848, opened 2024-04-11
Fix management edge mapping wrapper

SteveMonnier:


mPorhel:
@pdulth @SteveMonnier @ebausson-obeo  this seems like an automatic rebased PR, but the behavior has already been corrected and fixed by @pdulth in commit https://github.com/eclipse-capella/capella/commit/165318f5bd9da65faa62a395ebb60746a36aeeea

SteveMonnier:
Hi. My bad, I am not sure why but when I migrated to Capella 7.0 the sample for the validation scenario it somehow broke it. I migrated it again and it works with @pdulth commits. This PR can be abandonned.

PR #2885, opened 2024-07-05
[DRAFT] Improve DnD for DeploymentLinks

GlennPlou:


core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/FaServices.java
```diff
@@ -1,5653 +1,5658 @@
...
+  }
+
+  @Deprecated
+  public EdgeMapping getMappingFunctionalExchange(DDiagram diagram) {
+    String mappingName = MappingConstantsHelper.getMappingFunctionalExchange(diagram);
+    return DiagramServices.getDiagramServices().getEdgeMapping(diagram, mappingName);
+  }
+
+  @Deprecated
+  public NodeMapping getMappingABAbstractFunction(AbstractFunction functi
```
scosta-obeo, line 2993: Note that you can use `.toList()` instead of `collect(Collectors.toList())` if you don't need to modify it.

scosta-obeo, line 3017: Note that you can use .toList() instead of collect(Collectors.toList()) if you don't need to modify it.

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/ABServices.java
```diff
@@ -727,6 +727,11 @@ public boolean isValidDndABComponent(Part semanticObjectToDrop, EObject targetCo
     }
 
     context = CsServices.getService().getABTarget((DSemanticDecorator) targetContainerView);
+    final EObject semanticTarget = context; 
+    boolean isSourceAlreadyReployed = semanticObjectToDrop.getDeployingLinks().stream().anyMatch(link -> link.eContainer().equals(semanticTarget));
```
scosta-obeo, line 731: The variable `semanticTarget` is useless. Use `context` variable directly.

scosta-obeo (approved):

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/ABServices.java
GlennPlou: Seems weird I know, but i can't use the `context` variable here since it is not final, and i can't make it final.

scosta-obeo: Yes you can make it final by removing the useless initialization to null.

tguiu (approved):

PR #2895, opened 2024-09-09
#2894 Avoid potential deadlocks in refresh

mPorhel:
Change-Id: I653a0e579d58abf71f44ceb8eb85179cc07d5eb5

Issue #2894, opened 2024-09-09
DiagramServices::getEditPart(DDiagramElement) can cause deadlock


mPorhel:
In some situations, deadlock can occur during refresh of diagrams having the capability to display functional chains or physical paths: ie diagrams with Sirius post refresh extensions :
* org.polarsys.capella.core.sirius.analysis.refresh.extension.ComponentArchitectureBlankRefreshExtension.postRefresh(DDiagram)
* org.polarsys.capella.core.sirius.analysis.refresh.extension.DataFlowBlankRefreshExtension.postRefresh(DDiagram)
* org.polarsys.capella.core.sirius.analysis.refresh.extension.EntityArchitectureBlankRefreshExtension.postRefresh(DDiagram)

The detected performance issue or potential deadlock comes from the DiagramServices::getEditPart(DDiagramElement)  which uses an helper using Display.syncExec method to get the ActiveEditor. 
This also means that a refresh occuring on a non active editor will not refresh the edit parts.

Reproduction scenario exists with Team for Capella: 
* when a user has received remote changes on a locked diagram, if the automatic refresh is disabled, this diagram will be frozen and will need a manual refresh to display the received remote changes and be refresh accordingly to potential local changes.  
* with Capella 6.1 and 7.0 (an dI assume 5.2/6.0), with the _[LAB] [Build] All Components, Functions, CEs, FEs_ diagram from the IFE sample, I get a ~29s manual refresh for this particular scenario compared to an immediate (200ms to 1s regarding the versions) time for other refresh use cases of the same diagram
* analysis targets the DiagramService::getEditPart implementation

mPorhel:
With the provided PR, the issue is no more reproduced with the scenario we have with Team for Capella.
The particular refresh done on a frozen diagram comes back to the same timings than other refreshes.

PR #2907, opened 2024-10-15
#2894 Avoid potential deadlocks in refresh

GlennPlou:
This PR takes over the commit of PR #2895 by fixing tests.

### DIFF
--- a/core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/DiagramServices.java
+++ b/core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/DiagramServices.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2006, 2023 THALES GLOBAL SERVICES.
+ * Copyright (c) 2006, 2024 THALES GLOBAL SERVICES and others.
  * 
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -9,6 +9,7 @@
  * 
  * Contributors:
  *    Thales - initial API and implementation
+ *    Glenn Plouhinec, Maxime Porhel (Obeo) - Avoid potential deadlocks in refresh
  *******************************************************************************/
 package org.polarsys.capella.core.sirius.analysis;
 
@@ -44,7 +45,6 @@ import org.eclipse.sirius.business.api.session.Session;
 import org.eclipse.sirius.business.api.session.SessionManager;
 import org.eclipse.sirius.common.tools.api.interpreter.IInterpreter;
 import org.eclipse.sirius.common.tools.api.util.RefreshIdsHolder;
-import org.eclipse.sirius.common.ui.tools.api.util.EclipseUIUtil;
 import org.eclipse.sirius.diagram.AbstractDNode;
 import org.eclipse.sirius.diagram.DDiagram;
 import org.eclipse.sirius.diagram.DDiagramElement;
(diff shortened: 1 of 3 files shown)


core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/DiagramServices.java
```diff
@@ -2100,13 +2102,20 @@ public void refreshRepresentationOfEditor(IEditorPart editor) {
    * @return
    */
   public EditPart getEditPart(DDiagramElement diagramElement) {
-    IEditorPart editor = EclipseUIUtil.getActiveEditor();
-    if (editor instanceof DiagramEditor) {
-      Session session = new EObjectQuery(diagramElement).getSession();
+    DDiagram parentDiagram = diagramElement == null ? null : diagramElement.getParentDiagram();
+    Session session = parentDiagram == null ? null : new EObjectQuery(parentDiagram).getSession();
```
zgrtls, line 2106: If parentDiagram is null, we can return null immediately instead of doing additional computations

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/DiagramServices.java
```diff
@@ -2100,13 +2102,20 @@ public void refreshRepresentationOfEditor(IEditorPart editor) {
...
+    IEditingSession uiSession = SessionUIManager.INSTANCE.getUISession(session);
+
```
zgrtls, line 2108: if uiSession is null, we can return null immediately

zgrtls (commented):
Only a few minor code enhancements

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/DiagramServices.java
mPorhel: Each computation has a guard and having multiple `return null` in this method will reduce readability compared to the current state.

Do not hesitate to make this step on your side on top of this PR as if you which to go on this direction this would have been true on the previous code state

zgrtls (approved):

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/DiagramServices.java
tguiu: Please consider the "return early" pattern.

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/DiagramServices.java
tguiu: Please consider the "return early" pattern.

tguiu (changes requested):

core/plugins/org.polarsys.capella.core.sirius.analysis/src/org/polarsys/capella/core/sirius/analysis/DiagramServices.java
zgrtls: It is not a return early but now it is more readable. Looks good to go.

tguiu (approved):

Issue #2927, opened 2025-01-28
Avoid NPE in CapellaNavigatorContentProvider


GlennPlou:
In team context, a NPE may occurs when a shared project is imported in local client and override an existing project with opened session.

`org.polarsys.capella.core.platform.sirius.ui.navigator.viewer.CapellaNavigatorContentProvider.getChildren(Object)`
```java
          if (referencingModel instanceof CapellaModel) {
            Project object = ((CapellaModel) referencingModel).getProject(domain);
            if (object != null) {
              rootProject.add(object);
            }
            resourcesDone.add(object.eResource());
          }
```

PR #2928, opened 2025-01-28
#2927 Avoid NPE if project is null and simplify cast

GlennPlou:
project.eResource() was called even if project was null.


pdulth (approved):

pdulth:
Build is weird, i try to relaunch

Issue #2913, opened 2024-11-06
Changing navigation of associations in CDB diagram recreates them modifying original layout or hiding them in unsynchronized mode


scosta-obeo:
Step to reproduce:
* Create CDB Diagram
* Create 2 classes and an association between them
* Create a bendpoint on the association edge
* In the property view, make the association navigable in both directions (by checking the checkbox "Is Navigable" which is unchecked)
* **Expected**: the edge is always visible and it has the same bendpoints
* **Actuel**: the bendpoints are reset (synchronized mode) or the edge disappears (unsynchronized mode)

PR #2929, opened 2025-02-05
#2913 Add test validating fixes for AssociationCDBMigration

SteveMonnier:
Addition of a test that validates that the migration passes and that the edge with same source and target is still displayed on the migrated diagram.

Change-Id: I1bb8c8c685c01c294b8b2fee6ceec96bc4cc343d


GlennPlou (changes requested):
Add your test to a test suite

pdulth (changes requested):
RCPTT tests are no longer launched as unmaintanable.

I think this is already covered by the previous junit RelationStability no?
https://github.com/eclipse-capella/capella/blob/master/tests/plugins/org.polarsys.capella.test.migration.ju/src/org/polarsys/capella/test/migration/ju/testcases/basic/RelationStability.java#L420-L421

If not, can you update the junit test instead.

SteveMonnier:
Hi @pdulth, 
I ran the test RelationStability on my environment with this current PR and the test is all green. Is it possible to retrigger the job (I don't think I have the access for it, I don't see a retrigger action)?
I have created a JUnit to replace this RCPTT, but I am currently facing a "MetadataException: Wrong model version detected" issue I need to fix.

SteveMonnier:
I replaced the RCPTT test by a JUnit one

Issue #2954, opened 2026-01-05
Renaming a model breaks it in non-UTF8 systems


Kellindil:
- Go to the regional settings of windows and change the "language for non-unicode programs" to "simplified chinese"
- Start Capella
- Create a new project, name it 中文测试项目
- Rename the project to 更改名称后的项目 following the recommendations:
  - Rename the project itself
  - Then rename the .capella file
  - Finally rename the .aird file
- Open the project

The activity explorer opens an empty page, no content is visible under the project, and representations if there were any now show in the "invalid representations" category.

<img width="1000" height="591" alt="Image" src="https://github.com/user-attachments/assets/74f1abfa-aba1-4650-bbf6-4db53aa72673" />

Opening the ".capella" file in text mode, we can see that the "name" of the root element contains garbled characters - `name="鏇存敼鍚嶇О鍚庣殑椤圭洰"`

Kellindil:
I see that this has been merged, can it be closed?

PR #2955, opened 2026-01-06
#2954 Do not break project name

Kellindil:
This is a fix for #2954 .

The capella file's encoding is determined by its content (encoding="UTF-8" in the xml header). This is respected by EMF so the bytes will always have valid UTF-8 whatever the machine's encoding.

`new String(content.getBytes("UTF-8"))` explicitly breaks this valid String: it takes the content's bytes as UTF-8 but re-encodes them in the default machine encoding (e.g. GBK for simplified chinese) and puts these bytes encoded as GBK in the resulting file, which will then be read as UTF-8 because of the file header.


pdulth (approved):

pdulth:
I think I can do it at merge but you can do it if you have your setup
will wait for sonar fix / build ok

PR #2966, opened 2026-04-02
[releng][draft] capture the UI test environment from inside the product

cbrun:
## Status

**Draft / experiment**
The immediate goal is to get this testcase executed on the current Jenkins UI environment and inspect the console output. If that works, I plan to follow up with a separate change documenting the observed environment.

  ## Why

  Capella UI/layout tests are sensitive to their execution environment.

  When layout positions differ, the cause is often environmental rather than functional, for example:
  - different fonts
  - different resolution or DPI
  - different shell/workbench geometry
  - different X11 or window-manager setup

  That makes local reproduction and infrastructure changes harder than they should be.

  I want to capture what the current UI test environment actually looks like from inside a real test run so it can later be documented and used as a reference, especially for future machine/OS migrations.

  ## What this change does

  This PR adds a diagnostic UI testcase, `UIEnvironmentFingerprintTest`, to the platform UI suite.

  It prints a structured fingerprint of the environment to the console, including:
  - Java / Eclipse runtime info
  - display and monitor geometry
  - workbench shell bounds/state
  - SWT / JFace font resolution
  - selected environment variables
  - a few Linux/X11/font-related command outputs

  The test is observational: it records the environment, it does not try to lock it down with strict assertions.

  ## Intended next step

  If the output is useful on upstream Jenkins, I will submit a follow-up change to document the environment more explicitly.

  ## Review expectation

  No detailed review requested yet.
  This PR is mainly to validate the approach and collect the environment data.

### DIFF
--- a/tests/plugins/org.polarsys.capella.test.platform.ju/src/org/polarsys/capella/test/platform/ju/testcases/UIEnvironmentFingerprintTest.java
+++ b/tests/plugins/org.polarsys.capella.test.platform.ju/src/org/polarsys/capella/test/platform/ju/testcases/UIEnvironmentFingerprintTest.java
@@ -1,0 +1,279 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Obeo.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.polarsys.capella.test.platform.ju.testcases;
+
+import java.io.BufferedReader;
+import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


cbrun:
The `UIEnvironmentFingerprintTest` output on Jenkins gave us a few concrete facts that matter for reproducing the UI/layout tests reliably.

  - The Jenkins UI tests run on Linux/X11 with:
    - `1024x768`
    - `24-bit` depth
    - `96x96` DPI    
  - The workbench shell is **not maximized/fullscreen** and its client area is:
    - `1018x714`
  - The locale is:
    - `LANG=en_US.UTF-8`
  - Font resolution is a major point:
    - `fc-match sans` -> `DejaVu Sans`
    - `fc-match serif` -> `DejaVu Serif`
    - `fc-match monospace` -> `DejaVu Sans Mono`
    - `fc-match "Segoe UI"` -> `DejaVu Sans`
  - The test environment uses `Xvnc`/X11 tooling with `xrandr`, `xdpyinfo`, `xset` available.
  - The Jenkins runtime JVM is Java 17.

Issue #2967, opened 2026-04-04
Line ending policy causes false modified files after fresh clone on Linux/macOS (*.java eol=crlf + mixed blobs)


cbrun:
## Summary

On a fresh clone of `capella` (observed on Linux), `git status` reports hundreds of modified files immediately, without intentional edits.
The root cause appears to be an EOL normalization mismatch:
  - project `.gitattributes` enforces `*.java text eol=crlf`
  - some tracked Java files are stored in the index/history with mixed line endings
  - checkout/index normalization then makes Git report many files as modified

This creates noisy working trees for contributors and makes real code changes harder to review.

## Environment

- OS: Linux (reproduced on 2026-04-04)
- Git: (please fill with `git --version`)
- Repository state: freshly cloned, no manual edits before first `git status`

## Reproduction

1. Fresh clone:
 ```bash
  git clone <repo-url>
  cd capella
```

2. Check status:
```
git status --short
```
3. Inspect EOL metadata for a flagged Java file:
```
git ls-files --eol <path/to/file.java>
```

4. Compare semantic vs whitespace-only diff:
```
git diff --shortstat
git diff -w --numstat
```

## Observed behavior

- Initial git status showed ~543 modified files.
- git diff --shortstat showed large insert/delete counts.
- git diff -w --numstat returned 0 (indicating whitespace/EOL-only changes).
- git ls-files --eol showed entries like:
- i/mixed w/crlf attr/text eol=crlf for affected .java files.

Example affected files:

- tests/plugins/org.polarsys.capella.test.model.ju/src/org/polarsys/capella/test/model/ju/CheckInterModelInconsistencyDetectionTestCase.java
- tests/plugins/org.polarsys.capella.test.semantic.queries.ju/src/org/polarsys/capella/test/semantic/queries/ju/categoryregistry/testcases/
    CategoryRegistry_GatherAllCategories.java
- tests/plugins/org.polarsys.capella.test.semantic.queries.ju/src/org/polarsys/capella/test/semantic/queries/ju/categoryregistry/testcases/
    CategoryRegistry_GatherTechnicalCategories.java

## Why this matters

- New contributors may think they already changed many files.
- Large false-positive diffs hide real changes.
- Increased risk of accidental EOL-only commits and merge conflicts across OSes.

## References

- GitHub Docs: use .gitattributes to define shared line-ending behavior across contributors
    https://docs.github.com/en/get-started/git-basics/configuring-git-to-handle-line-endings
[...]

PR #2972, opened 2026-05-04
#2967 Relax Java line-ending policy

pdulth:
Fresh Linux and macOS clones start with hundreds of false modifications because Java files are forced to CRLF on checkout while many tracked Java blobs are stored with LF or mixed endings.

Remove the Java-specific CRLF rule and rely on the existing text=auto normalization policy so fresh clones stay clean without a repository-wide renormalization commit.


etraisnel2 (approved):
Looks good to me

PR #2977, opened 2026-05-05
#2950 #2959 #2962 #2970 #2960 Fix CVEs in multiple dependencies

Kellindil:
This fixes the CVEs described in #2950, #2959, #2962, #2970 and #2960.

I did not bump the version of Jetty (see ticket #2961) because this would require updating the eclipse platform version as well and this would be to big a change for a maintenance version. This has to come later on.

After building a Capella product with these commits and comparing the list of plugins in the end product with what was present in Capella 7.0.1, here is the exact list of changes these will introduce:

ch.qos.logback.classic					1.2.3	-> 1.5.32
ch.qos.logback.core					1.2.3	-> 1.5.32
ch.qos.logback.slf4j					1.2.3	-> removed
com.google.guava.failureaccess			1.0.1	-> 1.0.3
com.google.guava						21.0.0	-> removed
com.google.guava						31.1.0	-> 33.5.0
com.sun.jna.platform					5.12.1	-> 5.18.1
com.sun.jna							5.12.1	-> 5.18.1
com.sun.xml.bind						2.3.3	-> removed
jakarta.activation						1.2.1	-> removed
jakarta.xml.bind						2.3.3	-> removed
javaewah								1.1.13	-> com.googlecode.javaewah.JavaEWAH 1.2.3
javax.activation						2.0.0	-> removed
javax.xml								1.4.1	-> removed
org.apache.aries.spifly.dynamic.bundle		none	-> 1.3.7
org.apache.commons.commons-io		2.11.0	-> 2.21.0
org.apache.commons.httpclient			3.1.0	-> removed
org.apache.felix.scr						2.2.6	-> 2.2.14
org.apache.httpcomponents.httpclient		4.5.14.v20221207-10494	-> 4.5.14.v20230516-1249
org.apache.log4j						1.2.24	-> 1.2.26
org.apache.lucene.*					9.5.0	-> removed
org.apache.sshd.osgi					2.9.2	-> 2.10.0
org.apache.sshd.sftp					2.9.2	-> 2.10.0
org.apache.xerces						2.12.2	-> removed
org.apache.xml.resolver					1.2.0	-> removed
org.bouncycastle.bcpkix					none	-> 1.72.0
org.bouncycastle.bcutil					none	-> 1.72.0
org.eclipse.acceleo.*					7.0.0.202311201319	-> 7.0.0.202402190857
org.eclipse.egit.*						6.5.0	-> 6.6.1
org.eclipse.jgit.*						6.5.0	-> 6.6.1
[...]

PR #2982, opened 2026-05-12
Bump to 7.1.0

mPorhel:
- Bump commits can be merged after review and ci checks
- capella.vp modification, .afm expected version and modification of tests data might still needs discussion
- for the moment only Capella is bumped, Kitalpha will be bumped if needed.

### DIFF
--- a/common/features/org.polarsys.capella.common.feature/feature.xml
+++ b/common/features/org.polarsys.capella.common.feature/feature.xml
@@ -14,7 +14,7 @@
 <feature
       id="org.polarsys.capella.common.feature"
       label="%featureName"
-      version="7.0.1.qualifier"
+      version="7.1.0.qualifier"
       provider-name="%providerName">
 
    <description url="http://www.example.com/description">
--- a/common/features/org.polarsys.capella.common.feature/pom.xml
+++ b/common/features/org.polarsys.capella.common.feature/pom.xml
@@ -15,12 +15,12 @@
   <parent>
     <groupId>org.polarsys</groupId>
     <artifactId>org.polarsys.capella</artifactId>
-    <version>7.0.1-SNAPSHOT</version>
+    <version>7.1.0-SNAPSHOT</version>
     <relativePath>../../../</relativePath>
   </parent>
   <groupId>org.polarsys</groupId>
   <artifactId>org.polarsys.capella.common.feature</artifactId>
-  <version>7.0.1-SNAPSHOT</version>
+  <version>7.1.0-SNAPSHOT</version>
   <packaging>eclipse-feature</packaging>
   <properties>
     <!-- Skip feature modules in Sonar analysis -->
--- a/common/features/org.polarsys.capella.common.ui.feature/feature.xml
+++ b/common/features/org.polarsys.capella.common.ui.feature/feature.xml
@@ -12,7 +12,7 @@
 <feature
       id="org.polarsys.capella.common.ui.feature"
       label="%featureName"
-      version="7.0.1.qualifier"
+      version="7.1.0.qualifier"
       provider-name="%providerName">
 
    <description url="http://www.example.com/description">
--- a/common/features/org.polarsys.capella.common.ui.feature/pom.xml
+++ b/common/features/org.polarsys.capella.common.ui.feature/pom.xml
@@ -17,12 +17,12 @@
   <parent>
     <groupId>org.polarsys</groupId>
     <artifactId>org.polarsys.capella</artifactId>
-    <version>7.0.1-SNAPSHOT</version>
+    <version>7.1.0-SNAPSHOT</version>
     <relativePath>../../../</relativePath>
   </parent>
   <groupId>org.polarsys</groupId>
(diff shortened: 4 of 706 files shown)


pdulth:
Note for merger: 
To squash (and rename commit message to 7.1.0)

pdulth:
@mPorhel, maybe to merge after #2977.
Can you see with @Kellindil to see if we restore the reexport package as it has impact outside capella core product (capella addons, maybe some external addons too)

mPorhel:
I will update the PR today with corrected message I wrote 7.0.1 but read 7.1.0 ans pasted it several times.

I will also split the commits differently : I will keep capella.vp and .afm modifications for another PR we will wait to see if we decide to require migration of data or keep the current "model version".

There is at least one remaining topic : the splash screen which is still a 7.0.1 but in .bmp format.
There are some discussion about this on a dedicated ticket: #2983

mPorhel:
I have reworked the PR to squash releng commits and update the commit title.
There is a commit in my fork that I will propose for a further PR : https://github.com/mPorhel/capella/commit/56032c7792d151fb325a669d10ca5ff23dcbba7f

It contains version modification in .afm/.capella test models and the capella.vp bump. This would mean that model migration is required. For the moment it is not.

As non commiter, I cannot create a feature branch and stack the two PRs

To check: 
* org.polarsys.capella.test.migration.ju.fwk.MigrationRequiredTest
```
 assertTrue(requireMigration("7.0.2", "7.1.0"));
 ```

mPorhel:
> @mPorhel, maybe to merge after #2977. Can you see with @Kellindil to see if we restore the reexport package as it has impact outside capella core product (capella addons, maybe some external addons too)

@pdulth #2977 has been merged. With https://github.com/eclipse-capella/capella/issues/2950 [Switch back to reexport log4j on tools.report](https://github.com/eclipse-capella/capella/pull/2977/commits/c0ec66da51244d1c0329fcde19e877dbb99e73b1)

This PR cannot be built for the moment: it looks for library capella-pipeline-library@tag-nightly which is still commit a048167685c02842bd30fc223c16d5a498e079e9 "Fix JDK Download" and not yet commit e3814399ef9ebc9ca9d32e23457919168be69862 "[releng] Prepare the move to JDK 21: add fetch/retrieveTemurinJDK
"

mPorhel:
Tests were ok with this additional commit : https://github.com/eclipse-capella/capella/commit/ba418f29732333e1c7503840e2300b489deabe5b

We have to check what is this freeze in DiffMerge tests

Kellindil:
The diff merge tests freeze because they rely on the serialized comment in the ".capella" file to check if the models need to be migrated (see org.polarsys.capella.core.compare.CapellaScope.checkCapellaVersion(Resource)). This will look for the "\" comment in the capella file and parse it to find the major.minor.patch version and compare major.minor with the version of the capella feature currently installed in the running product. This does not fail in debug mode because Capella cannot detect the "version of the capella feature" in debug mode (the feature is not even there in that case). See also org.polarsys.capella.common.bundle.FeatureHelper.getCapellaVersion(boolean) and its comment "This method doesn't work on Debug mode."

We'll push a new commit on this PR to change the "Capella_Version" serialized in the tests files to 7.1.0.

Additional note : the way diff/merge checks for this version compatibility is extremely brittle : it assumes the version number can only be 5 characters long, textually parses something in the .capella file, and does not work in debug mode. This should most likely be analyzed to switch this flimsy "CapellaScope.checkCapellaVersion(Resource)" to the code that is used everywhere else to check for migration requirement (org.polarsys.capella.core.af.integration.CapellaMetadataProvider.isMigrationRequired(Version, Version) I believe).

pdulth:
It may be easier to do a classic bump updating all references from 7.0.0 and 7.0.1 to 7.1.0 except `/7.0.0` in all files. ([bump](https://github.com/eclipse-capella/capella/wiki/Development-How-To#bump-to-new-release)) 
And update the https://github.com/eclipse-capella/capella/blob/master/core/plugins/org.polarsys.capella.core.af.integration/src/org/polarsys/capella/core/af/integration/CapellaMetadataProvider.java#L193 to avoid migration from 7.0 to 7.1

Kellindil:
When started locally, on a windows machine, all of the "RelationStability" tests fails, and all of them keep the same number of bend points but have differing 'y' coordinates, for example : `junit.framework.AssertionFailedError: Wrong bendpoints after migration: expected:<{(58, 72), (120, 20), (170, 100), (260, 40), (300, 66)}> but was:<{(58, 75), (120, 22), (170, 102), (260, 42), (300, 69)}>`
This is not the same failure we see on the jenkins on which only one test fails, and it seems like all of the bend points except the first and last have been removed, not to mention the 'y' coordinate changing by a few pixels : `<{(58, 72), (120, 20), (170, 100), (260, 40), (300, 66)}> but was:<{(58, 70), (300, 70)}>`
Is there any way to understand exactly what the failure is here?

pdulth (approved):

Issue #2985, opened 2026-05-26
CapellaScope parses the comment in capella files to detect versions instead of using the afm


Kellindil:
During the resolution of #2982 we had tests freezing on the CI : they were prompting for migration because of the `CapellaScope.checkCapellaVersion(Resource))`.

CapellaScope looks for the `<!--Capella_Version_7.0.0-->` comment in the ".capella" files and parse it to find the major.minor.patch version.

This check for version compatibility is extremely brittle : it assumes the version number can only be 5 characters long, textually parses something in the .capella file, and does not work in debug mode.
`CapellaScope.checkCapellaVersion(Resource)` should be changed to use the code that is used everywhere else to check for migration requirement (`org.polarsys.capella.core.af.integration.CapellaMetadataProvider`).

pdulth:
Original usecase was when comparing two .capella when there is no .afm aside to know which version is coming from. (around 1.1.x, now there is an afm aside)

Some usecases that might still lookup for the Capella_Version_ tag:
- When comparing two .capella using git (compare with previous)
- When comparing two .capellafragments in the same workspace (so on stored in a sub folder of the main projects)
- When comparing two .capellafragments using git (compare with previous)

pdulth:
Git part:
https://github.com/eclipse-diffmerge/org.eclipse.emf.diffmerge.core/tree/master/plugins/org.eclipse.emf.diffmerge.connector.git/src/org/eclipse/emf/diffmerge/connector/git/ext

mPorhel:
Hi @pdulth, 

With commit [3cf4b7db2a3e841d8cc1c4f09933c422cf7b2694](https://github.com/eclipse-capella/capella/commit/3cf4b7db2a3e841d8cc1c4f09933c422cf7b2694) "[releng] Bump Capella_Version in .capella to 7.1.0" added to the bump PR to make the test pass, every project created with Capella 7.1.0 or saved with Capella 7.1.0 has the 7.1.0 version in the .capella xml comment.

For the upcoming 7.1.0 release, what should be the overall behavior around migration ?

- as before, Capella 7.1.0 should not be able to open a 7.0.1 version, until user has migrated the project? 
  - in this case modification of _org.polarsys.capella.core.af.integration/model/capella.vp_ is required (alongside .afm files of the repo and IFE sample)
- first version whose models (Capella only, no-addons) would be compatible with 2 versions 7.0.1 and 7.1.0.
[...]

PR #2990, opened 2026-06-15
[releng] Bump model version and test models to 7.1.0

mPorhel:
PR to test behavior of test and ci build with capella.vp/afm/models migrated to 7.1.0

Related to #2985


pdulth (approved):

pdulth:
Let me know if its ok to merge this one

mPorhel:
Tests are in the same state than master.

It is more up to you to choose if Capella should keep the previous behavior around model/afm/capella.vp all in sync. And migration required even if no real changes for now in the model.

With this commit, everything seems consistent. Otherwise there will only be the xml comment put with the xmlSave which will be in 7.1.0 after a save performed with Capella 7.1.0.

pdulth:
Let's go, 
will be consistent with the previous 6.0 > 6.1 and .afm will have the correct version, that will prevent inconsistency issues afterwards.

PR #2991, opened 2026-06-15
[releng] Add splash SVG source and BMP converter

mPorhel:
Keep the editable splash artwork in SVG form while still shipping the BMP variant expected by Eclipse. The checked-in splash.bmp uses a specific Windows 3.x 24-bit uncompressed layout, so regenerating it by hand is easy to get wrong.

Add the SVG source and a conversion script that derives the target bitmap format from the existing splash.bmp and reproduces it deterministically.


pdulth (approved):
- There was some weird path conversions in the current SVG, among

Image

Image

Image

Image

- Changed star color to capella website one #FBBF52
https://github.com/eclipse-capella/capella-website/blob/master/css/main.css#L52

See https://github.com/mPorhel/capella/pull/5

pdulth (changes requested):
See previous.
Can you merge on your MR [first ](https://github.com/mPorhel/capella/pull/5) 
 https://github.com/pdulth/capella/tree/splash

mPorhel:
Slight difference with generation from the latest svg: 
macOs on the left, Windows on the right: 

image

Version rendering.
macOS: Lato-Light is known
Windows: Lato-light unknown -> regular

mPorhel:
@pdulth do you have Lato-Ligh, Lato-Regular installed or do you only have Lato ?
On Steve's Windows laptop he has the same rendering than me.

We might provide a variant of the svg with "weight" rather than the PostScript name

pdulth:
I have installed https://fonts.google.com/specimen/Lato?preview.script=Latn
Regular and Light

In a new commit, i have switch to only "Lato" rather than both, but from what i see it's looks like your "light on mac" now

image

mPorhel:
Same in progress here but not pushed yet

mPorhel:
Hi @pdulth : generation is now stable with your env and ours.
Seems good to go.

pdulth (approved):

